Issue #5, opened 2021-06-11
api method not accusable through generated package


alakendu:
Current snappi generator incorporate this within __init__.py which is missing in openapiart.
```python
from .snappi import api
```
I think it will better to add that to directly access api method. Otherwise we need to extract package to get api.

PR #6, opened 2021-06-14
Add api in top_level_factories

alakendu:
This should address https://github.com/open-traffic-generator/openapiart/issues/5


ashutshkumr (approved):
pipeline was failing due to extension which detects changes in file. synced with main to resolve issues.

PR #19, opened 2021-06-24
Snappi changes sync

Rangababu-R:
place holder
need to test the code

### DIFF
--- a/openapiart/common.py
+++ b/openapiart/common.py
@@ -159,7 +159,7 @@ class OpenApiBase(object):
         """Deserialize a python object into the current object.
 
         If the input `serialized_object` does not match the current
-        snappi object an exception will be raised.
+        openapi object an exception will be raised.
 
         Args
         ----
@@ -496,7 +496,7 @@ class OpenApiObject(OpenApiBase, OpenApiValidator):
     
     def get(self, name, with_default=False):
         """ 
-        getattr for snappi object
+        getattr for openapi object
         """
         if self._properties.get(name) is not None:
             return self._properties[name]
--- a/openapiart/common.py
+++ b/openapiart/common.py
@@ -7,6 +7,7 @@ import urllib3
 import io
 import sys
 import time
+import re
 
 try:
     from typing import Union, Dict, List, Any, Literal
--- a/openapiart/common.py
+++ b/openapiart/common.py
@@ -180,7 +180,126 @@ class OpenApiBase(object):
         raise NotImplementedError()
 
 
-class OpenApiObject(OpenApiBase):
+class OpenApiValidator(object):
+
+    _MAC_REGEX = re.compile(
+        r"^([\da-fA-F]{2}[:]){5}[\da-fA-F]{2}$")
+    _IPV6_REP1 = re.compile(r"^:[\da-fA-F].+")
+    _IPV6_REP2 = re.compile(r".+[\da-fA-F]:$")
+    _IPV6_REP3 = re.compile(
+        r"^[\da-fA-F]{1,4}:" *7 + r"[\da-fA-F]{1,4}$")
+    _HEX_REGEX = re.compile(r"^0?x?[\da-fA-F]+$")
+
+    __slots__ = ()
+
+    def __init__(self) -> None:
+        pass
+
+    def validate_mac(self, mac):
+        if mac is None:
+            return False
+        if isinstance(mac, list):
+            return all([
+                True if self._MAC_REGEX.match(m) else False
+                for m in mac
+            ])
+        if self._MAC_REGEX.match(mac):
+            return True
+        return False
+
+    def validate_ipv4(self, ip):
+        if ip is None:
+            return False
+        try:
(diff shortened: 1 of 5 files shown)


openapiart/common.py
```diff
@@ -233,15 +360,18 @@ def _set_property(self, name, value, choice=None):
...
+            self._validate_types(key, value)
             if isinstance(value, (OpenApiObject, OpenApiIter)):
                 output[key] = value._encode()
-            else:
+            elif value is not None:
                 output[key] = value
         return output
 
     def _decode(self, obj):
         snappi_names = dir(self)
```
ashutshkumr, line 373: Rename `snappi_names`

openapiart/common.py
```diff
@@ -233,15 +360,18 @@ def _set_property(self, name, value, choice=None):
...
+        dtypes = [list, str, int, float, bool]
         for property_name, property_value in obj.items():
             if property_name in snappi_names:
```
ashutshkumr, line 376: Rename `snappi_names`

openapiart/common.py
```diff
@@ -250,19 +380,26 @@ def _decode(self, obj):
                         property_value = child[1](self, property_name)._decode(property_value)
                     else:
                         property_value = child[1]()._decode(property_value)
-                elif isinstance(property_value, list) and property_name in self._TYPES:
+                elif isinstance(property_value, list) and \
+                    property_name in self._TYPES and \
+                        self._TYPES[property_name]["type"] not in dtypes:
                     child = self._get_child_class(property_name, True)
                     snappi_list = child[0]()
```
ashutshkumr, line 387: rename `snappi_list`

openapiart/common.py
```diff
@@ -250,19 +380,26 @@ def _decode(self, obj):
...
                     for item in property_value:
                         item = child[1]()._decode(item)
                         snappi_list._items.append(item)
                     property_value = snappi_list
```
ashutshkumr, line 391: Rename `snappi_list`

openapiart/common.py
```diff
@@ -287,14 +424,91 @@ def __eq__(self, other):
...
+            self.types_validation(property_value, details["format"], msg)
+
+    def validate(self):
+        self._validate_required()
+        for key, value in self._properties.items():
+            self._validate_types(key, value)
+    
+    def getproperty(self, name, get_default=False):
+        """ 
+        getattr for snappi object
```
ashutshkumr, line 488: Rename `snappi` and `getproperty` -> `get`

openapiart/generator.py
```diff
@@ -696,31 +709,55 @@ def _get_description(self, yobject):
...
+            "boolean": "bool", "array": "list",
+            "number": "float", "float": "float",
+            "double": "float"
+        }
+        if yproperty["type"]  in data_type_map:
+            return data_type_map[yproperty["type"]]
+        else:
+            return yproperty["type"]
 
     def _get_snappi_types(self, yobject):
```
ashutshkumr, line 724: Rename `_get_snappi_types`

openapiart/generator.py
```diff
@@ -345,17 +345,24 @@ def _write_snappi_object(self, ref, choice_method_name=None):
                 self._write()
```
ashutshkumr, line 345: should be following snippet:


```python
            # write _TYPES definition
            # TODO: this func won't detect whether $ref for a given property is
            # a list because it relies on 'type' attribute to do so
            snappi_types = self._get_snappi_types(schema_object)
            if len(snappi_types) > 0:
                self._write(1, '_TYPES = {')
                for name, value in snappi_types:
                    if len(value) == 1:
                        self._write(2, "'%s': {'%s': %s}," % (
                            name, list(value.keys())[0], list(value.values())[0]
                        ))
                        continue
                    self._write(2, "'%s': %s" % (name, '{'))
                    for n, v in value.items():
                        if isinstance(v, list):
                            self._write(3, "'%s': [" % n)
                            for i in v:
                                self._write(4, "'%s'," % i)
                            self._write(3, "],")
                            continue
                        self._write(3, "'%s': %s," % (n, v))
                    self._write(2, "},")
                self._write(1, '} # type: Dict[str, str]')
                self._write()
            else:
                # TODO: provide empty types as workaround because deserializer
                # in snappicommon.py currently expects it
                self._write(1, '_TYPES = {} # type: Dict[str, str]')
                self._write()
```

ashutshkumr (changes requested):
- Replace all mentions of snappi
- `_get_property_param_string` needs to be updated with latest code (I couldn't add comment in offending lines)

Rangababu-R:
>     * `_get_property_param_string`

Andy had changed the logic, which would suffice the fix, I had added in snappi generator
**if property["type"] in ["number", "integer", "boolean"]:**
                        val = "=%s" % default if default == "None" else "=%s" % default
                    else:
                        val = "=%s" % default if default == "None" else "='%s'" % default

ajbalogh (approved):
lgtm

PR #27, opened 2021-07-07
Refactor protobuf enum generation

ajbalogh:
Reference: https://developers.google.com/protocol-buffers/docs/style#enums

### Issue
Generated protobuf enums do not follow the google developers style guide.

### Fix
```
# this openapi def 
        choice:
          type: string
          enum: [g_d, g_e]
          default: 'g_d'

# translates to this protobuf definition
  enum ChoiceEnum {
    CHOICE_UNSPECIFIED = 0;
    CHOICE_G_D = 1;
    CHOICE_G_E = 2;
  }
  optional ChoiceEnum choice = 4 [
    (fld_meta).default = "ChoiceEnum.CHOICE_G_D",
    (fld_meta).description = "Description missing in models"
  ];
```

### DIFF
--- a/openapiart/openapiartprotobuf.py
+++ b/openapiart/openapiartprotobuf.py
@@ -99,6 +99,7 @@ class OpenApiArtProtobuf(OpenApiArtPlugin):
         return self._custom_id
 
     def _write_header(self, info_object):
+        self._write("// {}".format(self._info))
         for line in self._license.split("\n"):
             self._write("// {}".format(line))
         self._write()
@@ -160,9 +161,11 @@ class OpenApiArtProtobuf(OpenApiArtPlugin):
                     if openapi_object["format"] == "binary":
                         return "bytes"
                 elif "enum" in openapi_object:
-                    enum_msg = self._camelcase("{}Enum".format(property_name))
+                    enum_msg = self._camelcase("{}".format(property_name))
                     self._write_enum_msg(enum_msg, openapi_object["enum"])
-                    return enum_msg
+                    # append "Enum" to differentiate type from other composite
+                    # type; this eventually will be stripped off
+                    return enum_msg + "Enum"
                 return "string"
             if type == "integer":
                 return "int32"
@@ -179,6 +182,23 @@ class OpenApiArtProtobuf(OpenApiArtPlugin):
             camel_case += "{}{}".format(piece[0].upper(), piece[1:])
         return camel_case
 
+    def _uppercase(self, value):
+        upper_case = ""
+        insert_underscore = False
+        
+        for c in value:
+            if c.isupper():
+                if insert_underscore:
+                    upper_case += "_" + c
+                    insert_underscore = False
+                else:
+                    upper_case += c
+            else:
+                upper_case += c.upper()
+                insert_underscore = True
+
+        return upper_case
+
     def _get_description(self, openapi_object):
         if "description" in openapi_object:
             return openapi_object["description"].replace("\n", "\\n").replace('"', "")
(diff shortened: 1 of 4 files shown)


openapiart/openapiartprotobuf.py
```diff
@@ -183,14 +186,15 @@ def _get_description(self, openapi_object):
             return "Description missing in models"
 
     def _write_enum_msg(self, enum_msg_name, enums):
-        enum_msg = enum_msg_name.split(".")
-        self._write("message {} {{".format(enum_msg[0]), indent=1)
-        self._write("enum Enum {", indent=2)
+        """Follow google developers style guide for enums
+        - reference: https://developers.google.com/protocol-buffers/docs/style#enums
+        """
+        self._write("enum {} {{".format(enum_msg_name.replace(".", "")), indent=1)
```
ashutshkumr, line 192: I think we'll need to go back to following inclusion of UNSPECIFIED.
Kudos for fixing names to use proper camel case.

```proto
message Choice {
    enum Enum {
        UNSPECIFIED = 0;
        LACP = 1;
        STATIC = 2;
    }
}

  optional Choice.Enum choice = 1 [
    (fld_meta).default = "Choice.Enum.LACP",
    (fld_meta).description = "The type of LAG protocol."
  ];

```

Otherwise, enum scoping rules will mess things up in cases where there are two enum definitions in a given message.

I confirmed the same and was getting the following message (when running against models/artifacts/openapi.yaml)

```
otg.proto:740:5: "UNSPECIFIED" is already defined in "otg.DeviceBgpv4".
```

because DeviceBgpv4 has two enums defined inside it, 
- AsType
- AsNumberWidth

ashutshkumr (changes requested):
I think we'll need to go back to following along with inclusion of UNSPECIFIED.
Kudos for fixing names to use proper camel case.

```proto
message Choice {
    enum Enum {
        UNSPECIFIED = 0;
        LACP = 1;
        STATIC = 2;
    }
}

  optional Choice.Enum choice = 1 [
    (fld_meta).default = "Choice.Enum.LACP",
    (fld_meta).description = "The type of LAG protocol."
  ];

```

Otherwise, enum scoping rules will mess things up in cases where there are two enum definitions in a given message.

I confirmed the same and was getting the following message (when running against models/artifacts/openapi.yaml)

```
otg.proto:740:5: "UNSPECIFIED" is already defined in "otg.DeviceBgpv4".
```

because DeviceBgpv4 has two enums defined inside it, 
- AsType
- AsNumberWidth

ajbalogh:
@ashutshkumr  thanks for the pointing out the enum scoping problem. 
I want to follow the style guide so I modified the generator to produce enums in line with its suggestions, namely staying away from encapsulating in a message but the change in this case is to prefix enum values with the property name.
An additional property in the config confirms that there is no enum scoping problem with this change.

ashutshkumr (approved):
@ajbalogh I've pushed following extra changes. Kindly review and let me know if something should be changed.
- Fixed https://github.com/open-traffic-generator/openapiart/issues/29
- Fixed https://github.com/open-traffic-generator/openapiart/issues/30
- Slightly modified enum generation (removed Enum suffix from enum name and fixed enum field prefix case)

```proto
// before
 enum ActorActivityEnum {
    ACTORACTIVITY_UNSPECIFIED = 0;
    ACTORACTIVITY_PASSIVE = 1;
    ACTORACTIVITY_ACTIVE = 2;
  }

// after
  enum ActorActivity {
    ACTOR_ACTIVITY_UNSPECIFIED = 0;
    ACTOR_ACTIVITY_PASSIVE = 1;
    ACTOR_ACTIVITY_ACTIVE = 2;
  }
```

PR #35, opened 2021-07-13
Support generated add method

### LINKED ISSUE #21: Revisit on append method behavior of objectIter
There is a behavior change in append method on Iter object and breaking the existing behavior in snappi.
currently commented the append method portion in generator.py and shall revisit on the same.

ajbalogh:
Fixes #21

Support an additional method for adding an item to a class that inherits OpenApiIter.
The add method will provide params for all properties that are correctly defaulted.
It will return a reference to the newly created object.

### DIFF
--- a/openapiart/common.py
+++ b/openapiart/common.py
@@ -76,11 +76,15 @@ class HttpTransport(object):
             self.logger.warning("Certificate verification is disabled")
         self._session = requests.Session()
 
-    def send_recv(self, method, relative_url, payload=None, return_object=None):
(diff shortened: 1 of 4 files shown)


ashutshkumr (dismissed):
In generated code, i'm seeing an extra comma in most init functions. Is this intended ?

```python
    def __init__(self, parent=None, choice=None, ):
        super(LagProtocol, self).__init__()
        self._parent = parent
        self._choice = choice
```

ashutshkumr (approved):

Issue #36, opened 2021-07-14
In otg.proto file, no option to return warnings or errors for SetConfigResponse and other StatusCode(s)


PrasenjitAdhikary:
error / warning are not included in following structs. Same issue for other response code as well.
message SetConfigResponse {
  message StatusCode200 {
  }
  message StatusCode400 {
  }
  message StatusCode500 {
  }
  oneof statuscode {
    StatusCode200 statuscode200 = 1;
    StatusCode400 statuscode400 = 2;
    StatusCode500 statuscode500 = 3;
  }
}

PR #39, opened 2021-07-14
Fix reponse message in generated protobuf file

ashutshkumr:
Fixes https://github.com/open-traffic-generator/openapiart/issues/36

- Generate messages for schemas in components/responses
- Convert field names to lower snake case for request/response messages
- Include black and flake8 as testing requirement

Please see attached old and new .proto for comparison.

[new.otg.proto.txt](https://github.com/open-traffic-generator/openapiart/files/6817509/new.otg.proto.txt)
[old.otg.proto.txt](https://github.com/open-traffic-generator/openapiart/files/6817511/old.otg.proto.txt)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ __pycache__
 /build
 /dist
 /env
+/.vscode
+/scratch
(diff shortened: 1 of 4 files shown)


ajbalogh (approved):
did a diff of the old and new files (thanks for adding those) and LGTM.

PR #59, opened 2021-08-09
OpenAPIArt go ux module generator

ajbalogh:
What: Generate an openapiart go ux module from a bundled openapi.yaml file.

The go ux module generator will wrap the protoc generated protobuf/grpc stubs with a fluent interface.


openapiart/openapiart.py
```diff
@@ -144,11 +138,45 @@ def _generate(self):
...
             ]
-            print("grpc_tools.protoc args: {}".format(" ".join(process_args)))
+            print("Generating python grpc stubs: {}".format(" ".join(process_args)))
+            process = subprocess.Popen(process_args, shell=False)
+            process.wait()
+        except Exception as e:
+            print("Bypassed creation of python stubs: {}".format(e))
+
+        try:
+            grpc_dir = os.path.normpath(os.path.join(self._output_dir, self._python_module_name))
```
alakendu, line 148: Somehow my system is require to add go_package in proto file. We need to modify .proto if this will happen in CICD as well.

```protobuf
option go_package = "github.com/protocolbuffers/protobuf/examples/go/tutorialpb";
```

Issue #77, opened 2021-08-17
Go sdk path issue


Rangababu-R:
When openapiart is installed via pip and as the package reside under site-packages. fetching dir name with __file__ causing to fetch the site-package path of openapiart and all the go sdk stuff is getting created under it.

Rangababu-R:
available in version 0.0.51

PR #78, opened 2021-08-17
Go sdk path fix

Rangababu-R:
When openapiart is installed via pip and as the package reside under site-packages. fetching dir name with __file__ causing to fetch the site-package path of openapiart and all the go sdk stuff is getting created under it.
#77


ajbalogh (approved):
lgtm

Issue #79, opened 2021-08-17
SetConfig response is always error in go sdk generated file


ANISH-GOTTAPU:
SetConfig response is always error in go sdk generated file
Responses need to be generated.
Responses need to be added to return values for generated rpc methods.

ajbalogh:
I would like to change the generation of proto responses to the following in order to be consistent with other messages.
The different status codes should be generated as a choice enum.

```proto
message StatusCode200 {
  bytes bytes = 1;
}
message StatusCode400 {
  ErrorDetails error_details = 1;
}
message StatusCode500 {
  Error error = 1;
}
message SetConfigResponse {
  message Choice {
    enum Enum {
      unspecified = 0;
      status_code_200 = 1;
      status_code_400 = 2;
      status_code_500 = 3;
    }
  }
  Choice.Enum choice = 1;
  optional StatusCode200 = 2;
  optional StatusCode400 = 3;
  optional StatusCode500 = 4;
}
```

ashutshkumr:
Agree with this approach.

ashutshkumr:
@ajbalogh removed oneof but not keeping choices. Might want to add them later.

Moreover,
- GRPC response stubs will be fixed (will return correct response apart from errors)
- Anish needs to fix in the same branch the pytest for streaming capture (since streaming response is removed)

PR #95, opened 2021-08-18
Go response

alakendu:
At the time of supporting HTTP Transport. I found that response should require. It looks like it already tracked
https://github.com/open-traffic-generator/openapiart/issues/79

Anyway, I have put my proposal and open a PR for it. This is one code snippet from snappi.go after code generation

```go
func (api *snappiApi) SetConfig(config Config) (*otg.ResponseWarning, error) {
	if api.HasHttpTransport() {
		return api.httpSetConfig(config)
	}
	if err := api.grpcConnect(); err != nil {
		return nil, err
	}
	request := otg.SetConfigRequest{Config: config.msg()}
	ctx, cancelFunc := context.WithTimeout(context.Background(), api.grpc.requestTimeout)
	defer cancelFunc()
	client, err := api.grpcClient.SetConfig(ctx, &request)
	if err != nil {
		return err
	}
	resp, _ := client.Recv()
	switch resp.Statuscode.(type) {
	case *otg.SetConfigResponse_StatusCode_200:
		return resp.GetStatusCode_200().ResponseWarning, nil

	case *otg.SetConfigResponse_StatusCode_400:
		dest := resp.GetStatusCode_400().ResponseError
		data, _ := yaml.Marshal(&dest)
		return nil, fmt.Errorf(string(data))

	case *otg.SetConfigResponse_StatusCode_500:
		dest := resp.GetStatusCode_500().ResponseError
		data, _ := yaml.Marshal(&dest)
		return nil, fmt.Errorf(string(data))

	default:
		return nil, fmt.Errorf(resp.String())
	}
}

func (api *snappiApi) httpSetConfig(config Config) (*otg.ResponseWarning, error) {
	rsp, err := api.httpSendRecv("config", config.Json(), "POST")
	if err != nil {
		return nil, err
	}

	bodyBytes, err := ioutil.ReadAll(rsp.Body)
	defer rsp.Body.Close()
	if err != nil {
		return nil, err
	}

	switch {
	case rsp.StatusCode == 200:
		var dest otg.ResponseWarning
		if err := json.Unmarshal(bodyBytes, &dest); err != nil {
			return nil, err
		}
		return &dest, nil

	case rsp.StatusCode == 400:
		var dest otg.ResponseError
		if err := json.Unmarshal(bodyBytes, &dest); err != nil {
			return nil, err
		}
		data, _ := yaml.Marshal(&dest)
		return nil, fmt.Errorf(string(data))

	case rsp.StatusCode == 500:
		var dest otg.ResponseError
		if err := json.Unmarshal(bodyBytes, &dest); err != nil {
			return nil, err
		}
		data, _ := yaml.Marshal(&dest)
		return nil, fmt.Errorf(string(data))

	default:
		return nil, fmt.Errorf(string(bodyBytes))
	}
}
```


ajbalogh (changes requested):
Nothing otg.<struct> should be returned. All internal pb structs need to be wrapped.
The call to process request..$ref needs to be copied and called for reponses..$ref.
You can do this or wait for it come as part of my efforts.

ashutshkumr:
@ajbalogh could you please address generating correct response types so @alakendu can use that response typein generated http/grpc stubs.

ajbalogh:
Generation of go response types has been added to openapiart v0.0.59

PR #103, opened 2021-08-20
Optional Go Sdk gen

Rangababu-R:
Sample modifications to generate python and go sdks optionally.
as proto file generation is common for both python and go and also it is dependent on go_sdk_dir
So, setting the paths was not changed but added defaults to it and proto file is generated when openapiart object is instantiated.
`
from openapiart import OpenApiArt
OpenApiArt(
        api_files=[
        os.path.join(os.path.dirname(__file__), "./openapiart/tests/api/info.yaml"),
        os.path.join(os.path.dirname(__file__), "./openapiart/tests/common/common.yaml"),
        os.path.join(os.path.dirname(__file__), "./openapiart/tests/api/api.yaml"),
    ],
    ).GeneratePythonSdk().GenerateGoSdk()`


ajbalogh:
I think it would be more explicit if python sdk options went in the GeneratePythonSdk method and go sdk options in the GenerateGoSdk method.

PR #105, opened 2021-08-20
adding go unit tests

ANISH-GOTTAPU:
adding go unit tests


do.py
```diff
@@ -147,6 +147,27 @@ def test():
         else:
             print("Coverage thresold[{0}] is achieved[{1}]".format(coverage_threshold, result))
 
+    go_coverage_threshold = 35
+    # TODO: not able to run the test from main directory
+    os.chdir("pkg")
+    try:
+        run(["go test ./... -v -coverprofile coverage.txt | tee coverage.out"])
+    except:
+        os.chdir("..")
```
ashutshkumr, line 156: do this as part of "finally". don't repeat twice

pkg/mock_server_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net"
 
+	"github.com/golang/protobuf/ptypes/empty"
 	. "github.com/open-traffic-generator/openapiart/pkg/sanity"
```
ashutshkumr, line 10: do not use "."

prepend "sanity." to all the structs/funcs that are imported from "sanity".

pkg/unit_test.go
```diff
@@ -0,0 +1,155 @@
+package openapiart_test
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"log"
+	"reflect"
+	"testing"
+
+	. "github.com/open-traffic-generator/openapiart/pkg"
```
ashutshkumr, line 10: . "github.com/open-traffic-generator/openapiart/pkg"

instead of ".", use "openapiart". this will require prepending some structs/funcs with openapiart.

ashutshkumr (changes requested):

ashutshkumr:
Remove auto-generated code from repo. e.g. all non-tests files inside pkg/*

do.py
ANISH-GOTTAPU: updated

pkg/mock_server_test.go
ANISH-GOTTAPU: updated

pkg/unit_test.go
ANISH-GOTTAPU: updated

PR #157, opened 2021-08-31
update default location

ANISH-GOTTAPU:


ajbalogh (approved):

openapiart/common.py
```diff
@@ -59,7 +59,11 @@ class send_recv method to communicate with a REST based server.
 class HttpTransport(object):
     def __init__(self, **kwargs):
         """Use args from api() method to instantiate an HTTP transport"""
-        self.location = kwargs["location"] if "location" in kwargs else "https://localhost"
+        self.location = (
+            kwargs["location"]
+            if "location" in kwargs and kwargs["location"] is not None
+            else "https://localhost:443"
```
ajbalogh, line 65: Try a test with localhost vs 127.0.0.1. 
In the past I have seen that localhost takes seconds to resolve vs 127.0.0.1 which is quicker.

ashutshkumr (approved):

openapiart/common.py
ashutshkumr: Per last discussion, localhost might resolve to ::1 in IPv6-only systems, so we should go with localhost for now.

Regarding slowness, we have tried this only on linux so far and no slowness seen. @ANISH-GOTTAPU please check on windows (will need to connect to locally deployed ixn-api server)

ashutshkumr: BTW, FYI, in negative scenarios, where localhost doesn't have any server running, it indeed takes a lot of time, but the delay there is not due to localhost name resolution.

PR #174, opened 2021-09-13
Validation and defaults

Rangababu-R:
Pending:
Hex Validation,
Slice Validation of Ipv4, ipv6, hex
int and float min, max validation


Rangababu-R:
> 
> 
> Pending:
> Hex Validation,
> Slice Validation of Ipv4, ipv6, hex
> int and float min, max validation

Above pending work is done need to add UT tests.
Build is failing at go coverage.

openapiart/common.go
```diff
@@ -126,11 +130,137 @@ type httpClient struct {
...
 
-func Validate() {
+func validationResult() error {
 	if len(validation) > 0 {
 		for _, item := range validation {
 			fmt.Println(item)
 		}
-		panic("validation errors")
+		validation = nil
+		return fmt.Errorf("validation errors")
```
ajbalogh, line 139: I think it would be better to join all validation items into a string and return that as the error as opposed to printing them out, allowing users to report the validation errors however they choose.

ajbalogh (changes requested):

ajbalogh (changes requested):
Need to increase go coverage or reduce the pass rate to fix the build.

openapiart/openapiartgo.py
```diff
@@ -600,20 +628,40 @@ def _write_interface(self, new):
             }}
 
             func (obj *{struct}) ToPbText() string {{
+                obj.Validate()
+                vErr := validationResult()
+                if vErr != nil {{
+                    panic(vErr)
```
Rangababu-R, line 634: @ajbalogh @ashutshkumr , is it okay to panic for all To<Pb, Json, YAML> methods with in the framework or do we have to return string and error?

ajbalogh: I like the panic as it allows for the return to be used directly in methods like fmt.Println().
If we change the To<Pb, Json, Yaml> methods to return a tuple then we should add the Stringer interface to allow for a non-validation dump method.

ajbalogh (approved):

Issue #186, opened 2021-09-23
missing args in _get_struct_field_type


alakendu:
**Version information**
0.0.11

**Describe the bug**
Getting exception

**To Reproduce**
Run latest model0.6.1

**Expected behavior**
build should complete without any error

**Additional context**
call that missing arg
Also add default check

alakendu:
fix should available in v0.1.13

PR #187, opened 2021-09-23
minor bugfix #186

alakendu:
This should address https://github.com/open-traffic-generator/openapiart/issues/186

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1175,7 +1175,7 @@ class OpenApiArtGo(OpenApiArtPlugin):
                     )
             default = property_schema.get("default")
             if default is not None:
-                if field.setter_method == None:
+                if field.setter_method is None:
                     print("Warnning: Default should not accept for this property ", property_name)
                 else:
                     if field.type == "number":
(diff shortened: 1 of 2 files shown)


openapiart/openapiartgo.py
```diff
@@ -1175,11 +1175,14 @@ def _build_setters_getters(self, fluent_new):
                     )
             default = property_schema.get("default")
             if default is not None:
-                if field.type == "number":
-                    default = float(default)
-                if field.type == "bool":
-                    default = str(default).lower()
-                field.default = default
+                if field.setter_method is None:
```
Rangababu-R, line 1178: if field.type is in self._oapi_go_types.values
then
set default else continue
check if this logic make sense

alakendu: This is good approach. But self._oapi_go_types does not contain all types say ([]string, []int32). I know this is kind of indirect. Otherwise I need to validate any missing type.

alakendu: Use self._oapi_go_types with lstrip

Rangababu-R (approved):

PR #189, opened 2021-09-28
UT fix

Rangababu-R:


do.py
```diff
@@ -13,18 +13,21 @@
...
 def on_x86():
-    arch = platform.uname().machine.lower()
-    return arch == "x86_64"
+    return arch() == "x86_64"
 
 
 def on_linux():
-    return sys.platform == "linux"
+    print("The platform is {}".format(sys.platform))
+    return "linux" in sys.platform
```
ashutshkumr, line 30: use `==`

do.py
```diff
@@ -150,22 +153,20 @@ def test():
...
-                "Go tests achieved {1}% which is less than Coverage thresold {0}%,".format(
-                    go_coverage_threshold, result))
-        else:
-            print(
-                "Go tests achieved {1}% ,Coverage thresold {0}%".format(
-                    go_coverage_threshold, result))
+    ret = subprocess.run("go test ./... -v -coverprofile coverage.txt".split(), capture_output=True)
+    print(ret.stdout.decode("utf-8"))
+    if b'FAIL' in ret.stdout:
+        raise Exception("Go Tests Failed")
```
ashutshkumr, line 159: This will always raise error even if there's only one TC that failed. We actually want to fail only when coverage report is below threshold.

do.py
```diff
@@ -150,22 +153,20 @@ def test():
...
-        result = re.findall(r"coverage:.*\s(\d+)", out)[0]
-        if int(result) < go_coverage_threshold:
-            raise Exception(
-                "Go tests achieved {1}% which is less than Coverage thresold {0}%,".format(
-                    go_coverage_threshold, result))
-        else:
-            print(
-                "Go tests achieved {1}% ,Coverage thresold {0}%".format(
-                    go_coverage_threshold, result))
+    ret = subprocess.run("go test ./... -v -coverprofile coverage.txt".split(), capture_output=True)
```
ashutshkumr, line 156: this call also needs to check for exit code (i.e. check=True). moreover, please ensure this works with python2

ashutshkumr (changes requested):

do.py
Rangababu-R: I had ran into a problem when the platform name was linux2, that's when I changed to in

do.py
Rangababu-R: yes, Ideally if one TC fails the CICD shall fail right. in the last CICD you noticed even though the testcases failed the coverage was beyond the threshold.

do.py
Rangababu-R: openapiart has python3 only that's when I had added subprocess.run, I have noticed "run" method isn't supported in python2

ashutshkumr: we need to reuse this in snappi repo as well (which needs to support python2)

do.py
ashutshkumr: then we need to fix the threshold value. otherwise there's no point in measuring coverage (because we'll fail in advance)

Rangababu-R: I think TC validation shall get high priority than coverage, if testcases are failing there is no point in checking the coverage

do.py
ashutshkumr: ok got it

do.py
ashutshkumr: I mean we can move exception checking part to end (i.e. after coverage check)

ashutshkumr (approved):

PR #190, opened 2021-10-01
Fix for int64 with str type from gosnappi json

Rangababu-R:
https://github.com/open-traffic-generator/snappi/issues/115

### DIFF
--- a/openapiart/common.py
+++ b/openapiart/common.py
@@ -258,16 +258,6 @@ class OpenApiValidator(object):
         if max is not None and value > max:
             return False
         return True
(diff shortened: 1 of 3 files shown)


ajbalogh (approved):

ashutshkumr:
@Rangababu-R please add unit tests for following scenarios:
- deserialze numeric value to int64
- deserialze string value to int64
- serialize int64 to string value

Ensure this unit test exists for both snappi and gosnappi

ashutshkumr (approved):

PR #211, opened 2021-10-21
Hex Value and FromJson defaults fixes

Rangababu-R:
https://github.com/open-traffic-generator/snappi/issues/125
https://github.com/open-traffic-generator/snappi/issues/124

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1017,7 +1017,11 @@ class OpenApiArtGo(OpenApiArtPlugin):
                 if string(value) != "{name}" {{
                     obj.obj.{external_name} = nil
                 }}
-            """.format(name=name, external_name=self._get_external_struct_name(name)) for name in field.enums]
+            """.format(
+                name=name,
+                external_name=self._get_external_field_name(name)
+                ) for name in field.enums
+            ]
             self._write(
                 """func (obj* {struct}) Set{fieldname}(value {interface}{fieldname}Enum) {interface} {{
                 intValue, ok := {pb_pkg_name}.{interface}_{fieldname}_Enum_value[string(value)]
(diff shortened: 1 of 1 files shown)


Rangababu-R:
> 
> 
> > > Another issue with enum I noticed is that in generated code:
> > > ```
> > > * we're not setting other enum choices to nil when setting a particular choice which we should
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```go
> > > // SetGE sets the float64 value in the GObject object
> > > //  description is TBD
> > > func (obj *gObject) SetGE(value float64) GObject {
> > > 	obj.obj.GE = &value
> > > 	obj.SetChoice(GObjectChoice.G_E)
> > > 	return obj
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The problem right now is, when executing following snippet:
> > > ```go
> > > 	api := openapiart.NewApi()
> > > 	eObject := api.NewPrefixConfig()
> > > 	eObject.RequiredObject()
> > > 	a := eObject.G().Add()
> > > 	a.SetGD("yo")
> > > 	a.SetGE(64.5)
> > > 
> > > 	fmt.Println(eObject.ToYaml())
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We see that value of both `GD` and `GE` is set, which is incorrect.
> > > ```yaml
> > > a: asdf
> > > b: 65
> > > c: 33
> > > g:
> > > - choice: g_e
> > >   g_a: asdf
> > >   g_b: 6
> > >   g_c: 77.7
> > >   g_d: yo
> > >   g_e: 64.5
> > >   g_f: a
> > > h: true
> > > required_object:
> > >   e_a: 1
> > >   e_b: 2
> > > response: status_200
> > > ```
> > 
> > 
> > same here, if there is way to set nil this issue will not be seen. when I had raised this upfront why we didn't take it seriously?
> 
> again, what's stopping us from setting `GD` and `GE` to `nil` ? I see those are optional attributes.
> 
> ```yaml
>     GObject:
>       x-include:
>       - '../common/common.yaml#/components/schemas/ListObject'
>       properties:
>         g_a:
>           type: string
>           default: asdf 
>         g_b:
>           type: integer
>           default: 6
>         g_c:
>           type: number
>           default: 77.7
>         choice:
>           type: string
>           enum: [g_d, g_e]
>           default: 'g_d'
>         g_d:
>           type: string
>           default: 'some string'
>         g_e:
>           type: number
>           format: double
>           default: 3.0
>         g_f:
>           description: |-
>             Another enum to test protbuf enum generation 
>           type: string
>           enum: [a, b, c]
>           default: a
> ```

are you not getting my point, go is not allowing to set nil directly. I had discussed this earlier

Issue #225, opened 2021-10-28
In SetChoice, no need add a check for assigning all properties to nil if actual property is set after calling SetChoice instead of before calling it.


ANISH-GOTTAPU:
**Version information**
v 0.1.21

**Describe the bug**
In SetChoice, no need add a check for assigning all properties to nil if actual property is set after calling SetChoice instead of before calling it.

**Expected behavior**
Existing:
```
func (obj *fObject) SetFA(value string) FObject {
	obj.obj.FA = &value
	obj.SetChoice(FObjectChoice.F_A)
	return obj
}

func (obj *fObject) SetChoice(value FObjectChoiceEnum) FObject {
	intValue, ok := sanity.FObject_Choice_Enum_value[string(value)]
	if !ok {
		validation = append(validation, fmt.Sprintf(
			"%s is not a valid choice on FObjectChoiceEnum", string(value)))
		return obj
	}
	enumValue := sanity.FObject_Choice_Enum(intValue)
	obj.obj.Choice = &enumValue

	if string(value) != "f_a" {
		obj.obj.FA = nil
	}

	if string(value) != "f_b" {
		obj.obj.FB = nil
	}

	return obj
}
```

Expected:
```
func (obj *fObject) SetFA(value string) FObject {
        obj.SetChoice(FObjectChoice.F_A)
	obj.obj.FA = &value
	return obj
}

func (obj *fObject) SetChoice(value FObjectChoiceEnum) FObject {
	intValue, ok := sanity.FObject_Choice_Enum_value[string(value)]
	if !ok {
		validation = append(validation, fmt.Sprintf(
			"%s is not a valid choice on FObjectChoiceEnum", string(value)))
		return obj
	}
	enumValue := sanity.FObject_Choice_Enum(intValue)
	obj.obj.Choice = &enumValue

        obj.obj.FA = nil

        obj.obj.FB = nil

	return obj
}
```

ANISH-GOTTAPU:
fixed with #231

PR #231, opened 2021-10-29
fix for issue 225

ANISH-GOTTAPU:
The PR is to fix issue #225


ashutshkumr:
@ANISH-GOTTAPU Over here, `obj.ob.Values = make([]string, 0)` may be redundant. Could you please check ?

```go
func (obj *patternIpv4PatternIpv4) SetValues(value []string) PatternIpv4PatternIpv4 {
	obj.SetChoice(PatternIpv4PatternIpv4Choice.VALUES)
	if obj.obj.Values == nil {
		obj.obj.Values = make([]string, 0)
	}
	obj.obj.Values = value

	return obj
}
```

ashutshkumr (approved):

Issue #230, opened 2021-10-29
gosnappi helper `Items()` for iterators needs to be more efficient


ashutshkumr:
**Version information**
https://github.com/open-traffic-generator/openapiart/releases/tag/v0.1.22

**Describe the bug**
As of today, everytime `.Items()` is called on an iterator, we create a list of object wrappers and return it.

```go
func (obj *metricsPortMetricIter) Items() []PortMetric {
	slice := []PortMetric{}
	for _, item := range obj.obj.obj.Ports {
		slice = append(slice, &portMetric{obj: item})
	}
	return slice
}
```

**Expected behavior**
Given the fact that this is the most common way for processing each item in the iterator, we should cache the wrapper list to avoid GC pressure and additional processing when `.Items()` is called.

alakendu:
Issue should fix in https://github.com/open-traffic-generator/openapiart/releases/tag/v0.1.25

PR #234, opened 2021-11-02
internal wrapper list for gosnappi helper Items()

alakendu:
This should address https://github.com/open-traffic-generator/openapiart/issues/230

We will incorporate another internal variable within our wrapper package. We will fill that variable when access Items(). So it will not process each time. 
```go
type metricsResponse struct {
	obj          *snappipb.MetricsResponse
	portMetrics  []PortMetric
	flowMetrics  []FlowMetric
	bgpv4Metrics []Bgpv4Metric
	bgpv6Metrics []Bgpv6Metric
	isisMetrics  []IsisMetric
}

func (obj *metricsResponsePortMetricIter) Items() []PortMetric {
	return obj.obj.portMetrics
}

func (obj *metricsResponsePortMetricIter) Add() PortMetric {
	newObj := &snappipb.PortMetric{}
	obj.obj.obj.PortMetrics = append(obj.obj.obj.PortMetrics, newObj)
	newLibObj := &portMetric{obj: newObj}
	newLibObj.setDefault()
	obj.obj.portMetrics = append(obj.obj.portMetrics, newLibObj)
	return newLibObj
}

func (obj *metricsResponsePortMetricIter) Append(newObj PortMetric) MetricsResponsePortMetricIter {
	obj.obj.obj.PortMetrics = append(obj.obj.obj.PortMetrics, newObj.Msg())
	obj.obj.portMetrics = append(obj.obj.portMetrics, newObj)
	return obj
}

func (obj *metricsResponsePortMetricIter) Set(index int, newObj PortMetric) MetricsResponsePortMetricIter {
	obj.obj.obj.PortMetrics[index] = newObj.Msg()
	obj.obj.portMetrics[index] = newObj
	return obj
}
```
We put entire data within internal autogenerated pb file. Within this fix we are putting those list data within gosnappi package.

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1183,12 +1183,6 @@ class OpenApiArtGo(OpenApiArtPlugin):
             }}
 
             func (obj *{internal_struct}) Items() {field_type} {{
-                if len(obj.obj.{internal_items_name}) != len(obj.obj.obj.{field_name}) {{
-                    obj.obj.{internal_items_name} = {field_type}{{}}
-                    for _, item := range obj.obj.obj.{field_name} {{
-                        obj.obj.{internal_items_name} = append(obj.obj.{internal_items_name}, &{field_internal_struct}{{obj: item}})
-                    }}
-                }}
                 return obj.obj.{internal_items_name}
             }}
 
@@ -1457,12 +1451,21 @@ class OpenApiArtGo(OpenApiArtPlugin):
                 if field.struct and field.isEnum is False:
                     statements.append(
                         """if obj.obj.{name} != nil {{
+                            if set_default {{
+                                obj.{internal_items_name} = {field_type}{{}}
+                                for _, item := range obj.obj.{name} {{
+                                    obj.{internal_items_name} = append(obj.{internal_items_name}, &{field_internal_struct}{{obj: item}})
+                                }}
+                            }}
                             for _, item := range obj.{name}().Items() {{
                                 item.validateObj(set_default)
                             }}
                         }}
                         """.format(
-                            name=field.name
+                            name=field.name,
+                            field_type=field.type,
+                            internal_items_name="{}s".format(field.struct),
+                            field_internal_struct=field.struct
                         )
                     )
                     valid += 1
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -715,9 +715,18 @@ class OpenApiArtGo(OpenApiArtPlugin):
             return
         else:
             new.generated = True
+
+        self._build_setters_getters(new)
+        internal_items = []
+        for field in new.interface_fields:
+            if field.adder_method is not None and field.isArray is True:
+                internal_items.append("{}s {}".format(
+                    field.struct, field.type
+                ))
         self._write(
             """type {struct} struct {{
                 obj *{pb_pkg_name}.{interface}
+                {internal_items}
             }}
             
             func New{interface}() {interface} {{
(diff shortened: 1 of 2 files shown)


ajbalogh (commented):
There is no validation test case that supports the change.

alakendu:
I think these UT indirectly covered those scenario:
```go
func TestIterAdd(t *testing.T) {
	api := openapiart.NewApi()
	config := api.NewPrefixConfig()
	config.G().Add()
	config.G().Add()
	assert.Equal(t, len(config.G().Items()), 2)
}

func TestIterAppend(t *testing.T) {
	api := openapiart.NewApi()
	config := api.NewPrefixConfig()
	config.G().Add()
	g := config.G().Append(openapiart.NewGObject())

	assert.Equal(t, len(g.Items()), 2)
}

func TestIterSet(t *testing.T) {
	api := openapiart.NewApi()
	config := api.NewPrefixConfig()
	name := "new name set on slice"
	config.G().Add().SetName("original name set on add")
	config.G().Add()
	g := config.G().Set(1, openapiart.NewGObject().SetName(name))

	assert.Equal(t, name, g.Items()[1].Name())
}
```

ashutshkumr (changes requested):
I see following snippet for `Items()`:

```go
func (obj *prefixConfigGObjectIter) Items() []GObject {
	if len(obj.obj.gObjects) != len(obj.obj.obj.G) {
		obj.obj.gObjects = []GObject{}
		for _, item := range obj.obj.obj.G {
			obj.obj.gObjects = append(obj.obj.gObjects, &gObject{obj: item})
		}
	}
	return obj.obj.gObjects
}
```

But nothing seems to have changed for `Add()` or `Append()` or `Set()`. What this means is `Items()` MUST up processing entire list every time the list is modified.

e.g. If somebody does a `Set()`, it does not change the length of the items, so `Items()` will return items that were cached previously.

Hence, here's what I think needs to be done:
- Wrapper list should point to each item in the internal list, but this should not be done as part of `Items()`
- `Items()` should only return wrapper list
- `Add()` and `Append()` should modify both wrapper and internal list (ensuring the both point to same newly added object)
- `Set()` might need to modify both wrapper and internal list as well, but I think it will be done a little differently

Regarding unit test, noting down some of the missing coverage:
- Set index 1 of list with 3 items, and ensure we get the same item upon `Items()`
- Set index n of list where n is invalid index
- When doing Add/Append, don't just check the length of list, also validate the contents

ashutshkumr:
FYI, this is what I'm seeing in the generated code:

```go
func (obj *prefixConfigGObjectIter) Items() []GObject {
	if len(obj.obj.gObjects) != len(obj.obj.obj.G) {
		obj.obj.gObjects = []GObject{}
		for _, item := range obj.obj.obj.G {
			obj.obj.gObjects = append(obj.obj.gObjects, &gObject{obj: item})
		}
	}
	return obj.obj.gObjects
}

func (obj *prefixConfigGObjectIter) Add() GObject {
	newObj := &sanity.GObject{}
	obj.obj.obj.G = append(obj.obj.obj.G, newObj)
	newLibObj := &gObject{obj: newObj}
	newLibObj.setDefault()
	return newLibObj
}

func (obj *prefixConfigGObjectIter) Append(newObj GObject) PrefixConfigGObjectIter {
	obj.obj.obj.G = append(obj.obj.obj.G, newObj.Msg())
	return obj
}

func (obj *prefixConfigGObjectIter) Set(index int, newObj GObject) PrefixConfigGObjectIter {
	obj.obj.obj.G[index] = newObj.Msg()
	return obj
}
```

alakendu:
Update UT accordingly

ashutshkumr:
Why do we enclose this block inside `set_defaults` flag (the snippet is form validateObj) ? We should initialize wrapper list regardless of whether set_default is true or false, right ?

```go
		if set_default {
			obj.gObjects = []GObject{}
			for _, item := range obj.obj.G {
				obj.gObjects = append(obj.gObjects, &gObject{obj: item})
			}
		}
```

PR #235, opened 2021-11-03
Exposing Grpc Connect and Client

Rangababu-R:
Addresses the below issue
https://github.com/open-traffic-generator/snappi/issues/134

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -496,7 +496,15 @@ class OpenApiArtGo(OpenApiArtPlugin):
                 req.Header.Set("Content-Type", "application/json")
                 req = req.WithContext(httpClient.ctx)
                 return httpClient.client.Do(req)
-            }}            
+            }}
+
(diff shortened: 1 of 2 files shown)


rudranil-das (approved):

Rangababu-R:
@ajbalogh Could you please provide your inputs on exposing grpcClient and Connect calls. As this will serve Athena-G implementation at controller

ajbalogh:
I still don't understand why this needs to be exposed - Please provide a fully coded use case.

Rangababu-R:
@rudranil-das @ashutshkumr Could you please provide the insights of implementation to Andy.

ajbalogh (commented):
This appears to be redundant as there already is a .SetConfig exposed.

ashutshkumr:
I agree, it's not clear to me why we need this.

ashutshkumr:
Checked with @rudranil-das further on the usecase. Existing functionality (`NewGrpcTransport()`) should suffice.

Issue #240, opened 2021-11-11
Validation doesn't happen for negative integer, negative float, float in min-max range 


ANISH-GOTTAPU:
**Version information**
0.1.24

**Describe the bug**
Validation is not happening for below 
1) Negative Integer
2) Negative Float
3) Float in min-max Range
4) List of Integers with negative value
5) List of Float with negative values

**To Reproduce**
```
func TestConfigGeneratedValidate(t *testing.T) {
	api := openapiart.NewApi()
	config := api.NewPrefixConfig()
	config.SetSpace1(-11).SetFullDuplex100Mb(-11).SetB(-100.11).SetC(-11).SetListOfIntegerValues([]int32{-1, -2, -3}).SetInteger64(-10000000000000000)
	v1 := config.RequiredObject()
	v1.SetEA(-100.11).SetEB(-1.7976931348623157e+308)
	assert.NotNil(t, v1.Validate())
```

**Expected behavior**
Validation should happen and errors need to be returned

Rangababu-R:
will be available in 0.1.26

Issue #239, opened 2021-11-11
Remove bool from Validate()


Rangababu-R:
**Version information**
0.1.24

**Describe the bug**
Remove defaults bool from Validate method on an interface

PR #241, opened 2021-11-11
Issue 240 and validate_method restructure

Rangababu-R:
Addressed #240 and https://github.com/open-traffic-generator/openapiart/issues/239
restructured the def _write_validate_method in an order of field type

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1521,7 +1521,7 @@ class OpenApiArtGo(OpenApiArtPlugin):
             body = """
             // {name} is required
             if obj.obj.{name}{enum} == {value} {{
-                validation = append(validation, "{name} is required field on {interface}")
+                validation = append(validation, "{name} is required field on interface {interface}")
             }} """.format(
                 name=field.name, interface=new.interface,
                 value=0 if field.isEnum and field.isArray is False else value,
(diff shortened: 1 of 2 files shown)


ashutshkumr (changes requested):
Why aren't we setting defaults when calling `Validate()` ?


```go
func (obj *prefixConfig) Validate() error {
	obj.validateObj(false)
	return validationResult()
}
```

ashutshkumr:
What's the motivation behind calling it `validateText` ? Shouldn't we just call `Validate()` ?

```go
err := obj.validateFromText()
```

ashutshkumr:
Following error message needs a bit of modifcation:

any int32 -> actual max value

```
PrefixConfig.Space_1 shall be in the range of [0, any int32] but Got
```

ashutshkumr:
Why are we checking whether `space_1` is less than 0 ? From spec, it seems as if it has no constraints:

```yaml
        space_1:
          type: integer
```

Rangababu-R:
> 
> 
> What's the motivation behind calling it `validateText` ? Shouldn't we just call `Validate()` ?
> 
> ```go
> err := obj.validateFromText()
> ```

After the unmarshal of a JSON, YAML or PBText we need to set the defaults as well. in the earlier approach we had used Validate with a defaults bool argument which would have control to set defaults or not, as we decided to remove the argument now separating the validate Api for FromJson(will be internal) and Validate for external.

ashutshkumr:
@Rangababu-R I'll hold the merge for now because I noticed the error messages were modified as well.

Although new error messages are better, this is breaking negative tests for controller (since they rely on older error message).

Let's use this PR to conclude on our final error messaging.

ashutshkumr (approved):

Issue #242, opened 2021-11-11
[go] Missing HasStatus... if result is of type []byte


jkristia:
Add following test, using existing SetConfigResponse model. Notice it fails to compile as HasStatusCode200 is not generated

```go
func TestConfigHas200Result(t *testing.T) {
	cfg := openapiart.NewSetConfigResponse()
	cfg.SetStatusCode200([]byte("anything"))
	assert.True(t, cfg.HasStatusCode200())
}
```

PR #243, opened 2021-11-11
issue #242. Missing HasStatus...

jkristia:
### Change

Check for `StatusCode` in field name in code block that generate the 'Has..." method

### Test

Add `TestConfigHas200Result`. If it compiles, then HasStatusCode200 is generated

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1402,7 +1402,7 @@ class OpenApiArtGo(OpenApiArtPlugin):
                             fieldstruct=fieldstruct,
                             desc=self._api.components[fieldstruct].description.lstrip("// ")
                         )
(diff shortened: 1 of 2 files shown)


openapiart/openapiartgo.py
```diff
@@ -1402,7 +1402,7 @@ def _build_setters_getters(self, fluent_new):
                             fieldstruct=fieldstruct,
                             desc=self._api.components[fieldstruct].description.lstrip("// ")
                         )
-            if field.isOptional and field.isPointer:
+            if field.isOptional and field.isPointer or "StatusCode" in field.name:
```
ajbalogh, line 1405: Also check fluent_new.isRpcResponse in combination with StatusCode since some fields may contain that string?

ajbalogh (approved):

PR #244, opened 2021-11-12
restrict length for format ipv4 ipv6 and mac

alakendu:
This should address https://github.com/open-traffic-generator/models/issues/191. 

This will break https://github.com/open-traffic-generator/models rpo. Please modify entire change within model before merge it.

### DIFF
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -49,6 +49,7 @@ class Bundler(object):
         self._content = {}
         self._includes = {}
         self._resolved = []
+        self._errors = []
         yaml.add_representer(Bundler.description, Bundler.literal_representer)
 
     def _get_parser(self, pattern):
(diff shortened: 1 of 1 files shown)


openapiart/bundler.py
```diff
@@ -316,8 +323,13 @@ def _apply_common_x_field_pattern_properties(self, schema, xpattern, format, pro
         if format is not None:
             schema["format"] = format
         if "length" in xpattern:
-            schema["minimum"] = 0
-            schema["maximum"] = 2 ** int(xpattern["length"]) - 1
+            if "format" in xpattern and xpattern["format"] in ["ipv4", "ipv6", "mac"]:
+                self._errors.append("property %s should not contain length as format set to %s" %(
```
ajbalogh, line 327: The error should be more specific to indicate all the keywords that are not allowed such as minLength, maxLength, minimum, maximum.

ashutshkumr:
@ANISH-GOTTAPU could you please sync this branch with main and ensure the newly added tests are not failing ?

ANISH-GOTTAPU:
Closing this and is addressed in #254

Issue #221, opened 2021-10-20
Make error messages consistent and uniform across python and go SDK


shramroy:
Observation from Snappi version: 0.6.5

Some error message I am mentioning here for example - 

1. '0 <= PatternFlowIpv4TosReliability.Value <= 1 but Got 2\nvalidation errors' 
  - suggested way - PatternFlowIpv4TosReliability.Value is not in range [0, 1]
2. 'PortName is required field on interface DeviceEthernet'
- Adding line number for which error is coming will be helpful for this type of scenarios.

PR #248, opened 2021-11-17
Error message change for validation

Rangababu-R:
Sample code for the error messages by adding path to it
```go

func (obj *prefixConfig) FromJson(value string) error {
	opts := protojson.UnmarshalOptions{
		AllowPartial:   true,
		DiscardUnknown: false,
	}
	uError := opts.Unmarshal([]byte(value), obj.Msg())
	if uError != nil {
		return fmt.Errorf("unmarshal error %s", strings.Replace(
			uError.Error(), "\u00a0", " ", -1)[7:])
	}
	err := obj.validateFromText("prefixConfig.FromJson()")
	if err != nil {
		return err
	}
	return nil
}

func (obj *prefixConfig) validateFromText(path string) error {
	obj.validateObj(true, path)
	return validationResult()
}

func (obj *prefixConfig) Validate() error {
	obj.validateObj(false, "prefixConfig.Validate()")
	return validationResult()
}

func (obj *prefixConfig) validateObj(set_default bool, path string) {
	if set_default {
		obj.setDefault()
	}

	// RequiredObject is required
	if obj.obj.RequiredObject == nil {
		validation = append(
			validation,
			fmt.Sprintf("PrefixConfig.required_object is a required field from path %s", path))
	}

	if obj.obj.RequiredObject != nil {
		obj.RequiredObject().validateObj(
			set_default, fmt.Sprintf("%s.RequiredObject", path))
	}

```


Rangababu-R:
addressed: https://github.com/open-traffic-generator/openapiart/issues/221

Rangababu-R:
merged into #281

Issue #247, opened 2021-11-16
FromJson() accepts empty string but does not initialize objects to defaults resulting in nil pointer error on accessing getter


ashutshkumr:
**Version information**
0.1.26

**Describe the bug**
FromJson() accepts empty string but does not initialize objects to defaults. This breaks objects containing choice because when accessing choice getter, none of the choices are set.

**To Reproduce**
This will raise nil pointer error.
```go
	r := api.NewMetricsRequest()
	r.FromJson("")
	fmt.Println(r.Choice())
```

**Expected behavior**
nil pointer error shouldn't be raised, in fact the object should be initialized to defaults.

Rangababu-R:
available in 0.1.31

Issue #246, opened 2021-11-15
getter StatusCode200() method returns a new wrapper object instance everytime


Rangababu-R:
**Version information**
0.1.25

**Describe the bug**
This bug is induced because of the recent modification on list Items() method and affected the APIs that return a object with list properties.

**To Reproduce**
Example:
In the below code snippets the line ```obj.StatusCode200().FromJson(string(bodyBytes))``` will create a new instance of MetricResponse and FromJson will process the items of wrapper list, which is fine till then, but when returning ```obj.StatusCode200()``` instead of using processed by FromJson object the method returns a new instance again. 

```
if resp.StatusCode == 200 {
		obj := api.NewGetMetricsResponse()
		if err := obj.StatusCode200().FromJson(string(bodyBytes)); err != nil {
			return nil, err
		}
		err := obj.Validate()
		if err != nil {
			return nil, err
		}
		return obj.StatusCode200(), nil
	}
```
```
func (obj *getMetricsResponse) StatusCode200() MetricsResponse {
	if obj.obj.StatusCode_200 == nil {

		obj.obj.StatusCode_200 = NewMetricsResponse().Msg()
	}
	return &metricsResponse{obj: obj.obj.StatusCode_200}
}
```
```
type metricsResponse struct {
	obj          *snappipb.MetricsResponse
	portMetrics  []PortMetric
	flowMetrics  []FlowMetric
	bgpv4Metrics []Bgpv4Metric
	bgpv6Metrics []Bgpv6Metric
	isisMetrics  []IsisMetric
}
```

**Expected behavior**
StatusCode200() which is a wrapper method shall return existing metricsResponse struct when it is not nil

**Additional context**
As of now this issue is fixed partially in #241 PR at the API level, returning the object instance that is processed but need to discuss and conclude on the getter method

Rangababu-R:
available in 0.1.31

PR #250, opened 2021-11-17
Addressing #246 and small change in Iters wrapper list

Rangababu-R:
#246 and #247
Change in Iters wrapper list

Need to add Tests for Items

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -760,7 +760,6 @@ class OpenApiArtGo(OpenApiArtPlugin):
             }}
 
             func (obj *{struct}) FromPbText(value string) error {{
-                if value == `""` {{value = ""}}
                 retObj := proto.UnmarshalText(value, obj.Msg())
                 if retObj != nil {{
                     return retObj
@@ -792,7 +791,7 @@ class OpenApiArtGo(OpenApiArtPlugin):
             }}
 
             func (obj *{struct}) FromYaml(value string) error {{
-                if value == "" || value == `""` {{value = "{{}}"}}
+                if value == "" {{value = "{{}}"}}
                 data, err := yaml.YAMLToJSON([]byte(value))
                 if err != nil {{
                     return err
@@ -836,7 +835,7 @@ class OpenApiArtGo(OpenApiArtPlugin):
                     AllowPartial: true,
                     DiscardUnknown: false,
                 }}
-                if value == "" || value == `""` {{value = "{{}}"}}
+                if value == "" {{value = "{{}}"}}
                 uError := opts.Unmarshal([]byte(value), obj.Msg())
                 if uError != nil {{
                     return fmt.Errorf("unmarshal error %s", strings.Replace(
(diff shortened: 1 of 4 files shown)


ajbalogh (approved):

ashutshkumr (changes requested):
We shouldn't have to check for this. 

```go
	if value == `""` {
		value = ""
	}
```

What I mean is, we should just check for empty string and change it to `{}` wherever relevant. We shouldn’t be checking for string containing two double-quotes.

Rangababu-R:
> We shouldn't have to check for this.
> 
> ```go
> 	if value == `""` {
> 		value = ""
> 	}
> ```
> 
> What I mean is, we should just check for empty string and change it to `{}` wherever relevant. We shouldn’t be checking for string containing two double-quotes.

removed

Rangababu-R:
mistakenly closed

ashutshkumr:
Although I'm pretty much convinced that this change is in a correct direction, there are few things we should look into and implement if it makes sense. Please note, more evaluation may be needed to ensure that suggestions below will actually help:
- First something cosmetic. What's wrong with just using actual member name for wrapper/holder attributes (instead of suffixing them with `Holder`) ? The reason I'm mentioning this is because we now have so many names that compilation is now getting slower. Having shorter names may help.
- Every time we do `FromJson`, we should set all wrapper attributes to `nil`. Doing this ensures that underlying proto attribute is changed only from three methods:
  - `SetAttr()`
  - `Attr()` (upon calling `NewObj()` or `FromJson`, or in isolation)

  Thus hopefully this will simplify the logic of `Attr()`
  ```go
	func (obj *prefixConfig) RequiredObject() EObject {
		if obj.obj.RequiredObject == nil {
			obj.obj.RequiredObject = NewEObject().Msg()
		}

		if obj.requiredObjectHolder == nil {
			obj.requiredObjectHolder = &eObject{obj: obj.obj.RequiredObject}
		}

		return obj.requiredObjectHolder
	}
  ```

- Explore pointer-to-pointer for wrapper attributes. It requires more experiments, but will greatly reduce the work required to maintain wrapper objs. e.g.

```go
type eObject struct {
	obj **sanity.EObject
}

type prefixConfig struct {
	obj                   *sanity.PrefixConfig
	requiredObjectHolder  EObject
}

func NewPrefixConfig() PrefixConfig {
	obj := prefixConfig{obj: &sanity.PrefixConfig{}}
        obj.requiredObjectHolder = &eObject{obj: &obj.obj.RequiredObject}
	obj.setDefault()
	return &obj
}

func (obj *prefixConfig) RequiredObject() EObject {
	if obj.obj.RequiredObject == nil {
		obj.obj.RequiredObject = NewEObject().Msg()
	}

	return obj.requiredObjectHolder
}
```

- Need to understand changes related to iterators. The main reason why we got to this PR is because we were trying to fix an issue with iterators which in turn was a result of introducing wrappers for internal slices (to avoid extra work). But now that we see, we still end up doing similar amount of work. What's important is to check if the suggestions before this point make sense and can be implemented. If so, will it help reducing the work we do for iterators ?

ashutshkumr (approved):

PR #251, opened 2021-11-17
golangci-lint

ANISH-GOTTAPU:
golangci-lint


ANISH-GOTTAPU:
@ashutshkumr : Please check if this is ok for goLint which is added as part of do.py. If this is ok, I will clear the linting issues

ashutshkumr:
Can also install like so:

```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
```

If while installing, PWD doesn't contain go.mod, then

```
GO111MODULE=on go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
```

ANISH-GOTTAPU:
@ashutshkumr : getting this error with go install

package github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0: cannot use path@version syntax in GOPATH mode, any idea how to resolve this

https://github.com/open-traffic-generator/openapiart/runs/4300093500?check_suite_focus=true

ashutshkumr (approved):

Issue #252, opened 2021-11-18
route should created according to the valid HTTP methods


alakendu:
**Version information**
This is within branch https://github.com/open-traffic-generator/openapiart/tree/servergen/openapiart

**Describe the bug**
Getting this error when execute this PR to generate snappi
```
Traceback (most recent call last):
  File "D:\OTG\Codebase\snappi\setup.py", line 30, in <module>
    openapiart.OpenApiArt(
  File "D:\OTG\Codebase\VenvPython39\lib\site-packages\openapiart\openapiart.py", line 221, in GenerateGoServer
    servergen.generate()
  File "D:\OTG\Codebase\VenvPython39\lib\site-packages\openapiart\goserver\goserver.py", line 38, in generate
    self._loadyaml()
  File "D:\OTG\Codebase\VenvPython39\lib\site-packages\openapiart\goserver\goserver.py", line 56, in _loadyaml
    self._loadroute(url, pathobj)
  File "D:\OTG\Codebase\VenvPython39\lib\site-packages\openapiart\goserver\goserver.py", line 75, in _loadroute
    raise AttributeError(f"controller name missing from '{url} - {methodname}:'\nUse tags: [<name>]")
AttributeError: controller name missing from '/results/metrics - description:'
Use tags: [<name>]
```

**To Reproduce**
This is happening because it is expecting "description" should contain "tags"
```
  /results/metrics:
    description: |-
      Metrics API
    post:
      tags:
      - Metrics
```

**Expected behavior**
snappi should generate without any error

**Additional context**
Nil

alakendu:
@jkristia Could you please comment on methodobj restriction

jkristia:
Currently all methods on a given route must be on the same controller, and therefore the Tag (which is the controller name) must be at the route level, not method level.

jkristia:
yes this is a problem in the current implementation. It is expected to only contain valid http methods on the route. This will need to be fixed

alakendu:
This should available in https://github.com/open-traffic-generator/openapiart/releases/tag/v0.1.33

PR #253, opened 2021-11-18
loadroute only for valid http method

alakendu:
This should address https://github.com/open-traffic-generator/openapiart/issues/252


ajbalogh (commented):
I disagree with removing 3.x type annotations, type hints, or parameter type declarations.
I've created a draft PR to separate build from test so that using openapiart is done with 3.x versions while testing the generated artifacts can be done with 2.7 and 3.x matrix.

https://github.com/open-traffic-generator/openapiart/pull/269

Rangababu-R (approved):

Issue #262, opened 2021-11-30
Remove Panic from ToJson(), ToYaml() & ToPbText()


ANISH-GOTTAPU:
Remove Panic from ToJson(), ToYaml() & ToPbText()

ANISH-GOTTAPU:
Implemented with #267

PR #267, opened 2021-11-30
Removed Panic from generator and updated tests

ANISH-GOTTAPU:
To address #262 

Remove Panic from ToJson(), ToYaml() & ToPbText()


Rangababu-R:
For all ToJson, ToYaml and ToPbText calls in tests err shall be asserted on nil check
yaml1, _ := c1.ToYaml() --> current code
instead
yaml1, err := c1.ToYaml()
assert.Nil(t, err)

pkg/mock_httpserver_test.go
```diff
@@ -49,31 +49,38 @@ func StartMockHttpServer() {
 			case PrefixConfigResponse.STATUS_400:
 				response.StatusCode400().SetErrors([]string{"A 400 error has occurred"})
 				w.WriteHeader(http.StatusBadRequest)
-				w.Write([]byte(response.StatusCode400().ToJson()))
+				resp400, _ := response.StatusCode400().ToJson()
```
Rangababu-R, line 52: resp400, err := response.StatusCode400().ToJson()
add error check also and if there is an error server shall return 400 status object ji

Rangababu-R (changes requested):

pkg/mock_httpserver_test.go
ANISH-GOTTAPU: updated as discussed

Rangababu-R (approved):

openapiart/common.go
```diff
@@ -37,21 +37,21 @@ func (obj *grpcTransport) SetLocation(value string) GrpcTransport {
...
-	return int(obj.requestTimeout / time.Second)
+func (obj *grpcTransport) RequestTimeout() time.Duration {
+	return obj.requestTimeout / time.Second
 }
 
 // SetRequestTimeout contains the timeout value in seconds for a grpc request
-func (obj *grpcTransport) SetRequestTimeout(value int) GrpcTransport {
-	obj.requestTimeout = time.Duration(value) * time.Second
+func (obj *grpcTransport) SetRequestTimeout(value time.Duration) GrpcTransport {
+	obj.requestTimeout = value * time.Second
```
ashutshkumr, line 46: it should just accept value

openapiart/common.go
```diff
@@ -37,21 +37,21 @@ func (obj *grpcTransport) SetLocation(value string) GrpcTransport {
...
 	return obj
 }
-func (obj *grpcTransport) DialTimeout() int {
-	return int(obj.dialTimeout / time.Second)
+func (obj *grpcTransport) DialTimeout() time.Duration {
+	return obj.dialTimeout / time.Second
```
ashutshkumr, line 50: should just return timeout, without dividing it by seconds

ashutshkumr (changes requested):
Please review changes related to duration. I don't think multiplication or division with seconds should be involved.

ashutshkumr (approved):

Issue #277, opened 2021-12-10
Validation errors are not clubbed


Rangababu-R:
**Is your feature request related to a use case? Please describe.**
Validation error is being thrown when ever the first one is hit.
All the validation error shall be clubbed and error shall be raised at the end of the validation

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #276, opened 2021-12-10
Object type, required property is not being validated


Rangababu-R:
**Version information**
0.1.31

**Describe the bug**
The Object type property is not being validated, instead the default object is being created

**To Reproduce**
do not initialize the object property and try to validate, the required property will not raise a validation error 

**Expected behavior**
an error needs to be raised for the object type property.

**Additional context**
As there was no setter method in snappi the validation is creating the new object for the object typed properties.

Vibaswan:
This feature is already taken care of in a previous merge 
#433

Issue #274, opened 2021-12-10
Enum Setter and Getter behaviour is incorrect for invalid choices


Rangababu-R:
**Version information**
0.1.31

**Describe the bug**
snappi: enum setter accepts any value and the same can be retrieved in getter method.
gosnappi: Any invalid value can will be set to unspecified or default

**To Reproduce**
```
python:
c = sanity.Api().prefix_config()
c.response = "abc"
c.response returns abc
go:
c := openapiart.NewPrefixConfig()
c.SetResponse("abc")
c.Response() returns "status_200" as the default is status_200
```

**Expected behavior**
shall append to the validation errors, in setter

Issue #273, opened 2021-12-10
Object cloning method is unavailabe in gosnappi


Rangababu-R:
**Is your feature request related to a use case? Please describe.**
An Object can be cloned in snappi but not in gosnappi

**Describe the solution you'd like**
Need to add clone method to type object

PR #281, opened 2021-12-15
Py go diff

Rangababu-R:
- resolves https://github.com/open-traffic-generator/openapiart/issues/272
- resolves https://github.com/open-traffic-generator/openapiart/issues/274
- resolves https://github.com/open-traffic-generator/openapiart/issues/276
- resolves https://github.com/open-traffic-generator/openapiart/issues/277
- resolves https://github.com/open-traffic-generator/openapiart/issues/273
- resolves https://github.com/open-traffic-generator/openapiart/issues/221


openapiart/common.py
```diff
@@ -625,6 +670,12 @@ def append(self, item):
     def clear(self):
         del self._items[:]
         self._index = -1
+    
+    def set(self, index, item):
+        if isinstance(item, OpenApiObject) is False:
```
alakendu, line 675: Validate if item is really that iterobj. Otherwise it use will do this
```python
pc = api.prefix_config()
pc.deserialize(payload)
f = pc.f
j = pc.j
j.set(0, f)    # add validation
```

openapiart/common.py
```diff
@@ -398,8 +419,11 @@ def _set_property(self, name, value, choice=None):
             self._set_choice(name)
             self._properties[name] = self._DEFAULTS[name]
         else:
-            self._set_choice(name)
-            self._properties[name] = value
+            if not self._is_enum_valid(name, value):
```
alakendu, line 422: Not sure what is going wrong. But I am not getting any error whey try this
```python
pc = api.prefix_config()
pc.deserialize(payload)
pc.response = "abc"
pc.validate(True)    # return black though I am expecting enum error
```

openapiart/common.py
```diff
@@ -398,8 +419,11 @@ def _set_property(self, name, value, choice=None):
...
+                self._append_error("{} is not a valid enum for property {}".format(value, name))
+            else:
+                self._set_choice(name)
+                self._properties[name] = value
         if self._parent is not None and self._choice is not None and value is not None:
```
alakendu, line 427: though I have no ready use case. But I feel we need to add another condition when self._is_enum_valid() will return True. Otherwise we will silently modify that enum. Please ignore this comment if existing condition is sufficient.

openapiart/common.py
Rangababu-R: pc.validate(True) will skip raising the exception

openapiart/common.py
Rangababu-R: Agreed, I had the same concern while implementing. But for "append" call too the logic is same, so I followed it. will add the validation

openapiart/common.py
Rangababu-R: I feel the existing logic will suffice

ajbalogh (approved):
LGTM

ashutshkumr:
I would suggest some error message changes in order to ensure that both python and go refer to same `names` in the data models.

1. Do not expose internal names in error message:

    ```
    PrefixConfig.required_object is a required field from path prefixConfig.Validate()
    ```

    Recommended:
    ```
    required field `self.required_object` must not be empty
    ```

2. Rephrasing error message for length

    ```
    length of PrefixConfig.str_len shall be in the range of [3, 6] but Got 2 from path prefixConfig.Validate()
    ```

    Recommended:
    ```
    length of field `self.str_len` must be in range [3, 6], instead of 2
    ```

> Will add more points as I find them (w.r.t. error messages)

ashutshkumr:
Following snippet runs without issues, even though I did not set `EB` (required field) in `ja`.

```go
func TestNewAndSet(t *testing.T) {
	c := openapiart.NewPrefixConfig()
	c.RequiredObject()
	c.SetA("hello")
	ja := c.J().Add().JA()
	ja.SetEA(2.4)
	c.SetE(openapiart.NewEObject().SetEA(123.456).SetEB(453.123))
	c.SetF(openapiart.NewFObject().SetFA("fa string"))
	yaml1, err := c.E().ToYaml()
	assert.Nil(t, err)
	yaml2, err := c.F().ToYaml()
	assert.Nil(t, err)
	log.Println(yaml1)
	log.Println(yaml2)
	log.Println(c.ToYaml())
}
```

ashutshkumr:
For `Clone`, try converting to and from `PbText` instead of JSON since former would be faster (since our core data structure is in protobuf and not JSON)

ashutshkumr:
shall be merged as part of https://github.com/open-traffic-generator/openapiart/pull/320

PR #283, opened 2022-01-04
Support for stringer fuction

Rangababu-R:
fmt.Println(Obj) -> will print YAML string


ashutshkumr (approved):
LGTM, please sync with main after pipeline changes have gone in and fix linter issue too.

PR #286, opened 2022-01-08
Perf Scripts for python and go

Rangababu-R:
Will add the gold values once the flow is approved


ashutshkumr (approved):
@Rangababu-R The test looks good to me in terms of functionality. But we need to ENH it in following areas:
- instead of logging the time, output it in a JSON
- the current JSON should be compared with last available JSON (i.e. gold values) with some tolerance value
- if no previous JSON exists or the current value is better than last value, current JSON must be checked in
- if current value is worse than last value, test MUST fail
- We need to have a tolerance value when comparing measurements

PR #269, opened 2021-12-07
Separate build and test

ajbalogh:
### Issue
The CICD pipleline of openapiart is done using 2.7 and 3.x which hampers the coding effort as it must accommodate the lowest common version (2.7) which is currently EOL.

### Fix
Breakup the CICD workspace action into build (3.x), python test (2.7 - 3.x), go test and publish python package. 
This PR can also serve as a template for other pipelines that need 2.7/3.x artifacts.

### Validation
The CICD workspace action successfully passes.

PR #287, opened 2022-01-10
separate build with python3 then run test with python2 

alakendu:
This PR is top of https://github.com/open-traffic-generator/openapiart/pull/269 as well as sync with main and file change to Python3


.github/workflows/build.yml
```diff
@@ -1,58 +1,74 @@
...
         uses: actions/checkout@v2
         with:
           ref: ${{ github.head_ref }}
-      - name: Setup Go for CI
-        uses: actions/setup-go@v2
+      - uses: actions/setup-go@v2
         with:
-          go-version: '1.17.3'
-      - name: Setup Go and protoc
+          go-version: "^1.17.1"
```
ashutshkumr, line 26: let's use a fixed version of go (1.17.5)

.github/workflows/build.yml
```diff
@@ -1,58 +1,74 @@
...
+      - name: Setup protoc
         run: |
-          python do.py setup_ext
+          python do.py get_protoc
```
ashutshkumr, line 29: use `python setup_ext` only (since it checks for incorrect platform)

.github/workflows/build.yml
```diff
@@ -61,28 +76,54 @@ jobs:
...
+      - name: Display structure of downloaded files
+        run: ls -R
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.5"
+      - name: Setup protoc
         run: |
-          python do.py go_lint
-      - name: Run tests
+          python do.py setup_ext
```
ashutshkumr, line 106: Do we really need to do this ?

.github/workflows/build.yml
```diff
@@ -61,28 +76,54 @@ jobs:
...
+        uses: actions/checkout@v2
         with:
-          name: generated-artifacts
-          path: art
+          ref: ${{ github.head_ref }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Generate only for 3.9
```
ashutshkumr, line 123: can't we just download artifacts and post them (since that's the one we've tested) ?

do.py
```diff
@@ -132,16 +132,17 @@ def generate():
     )
 
 
-def test():
+def test_python():
```
ashutshkumr, line 135: i would suggest shortening the name to `testpy` and `testgo`.

.github/workflows/build.yml
alakendu: set to 1.17.5

.github/workflows/build.yml
alakendu: change setup_ext

.github/workflows/build.yml
alakendu: agrre

ashutshkumr (approved):

Issue #280, opened 2021-12-15
Add support to customize generated proto API with flexibility of service name, message start value and stream support


PrasenjitAdhikary:
1.	Ability to change service name from “Openapi” to custom name.
2.	Ability to change message prefix from “OpenApi” to custom name (or auto-generated from service name).
3.	Ability to start message starting value from 60001 to custom start value.
4.	Ability to specify “stream” in proto API.

ajbalogh:
@PrasenjitAdhikary in the original issue please edit the content to include examples of what the generated code would look like - it doesn't need to be the entire generated file, just relevant snippets.

PR #289, opened 2022-01-17
Customize proto

Rangababu-R:
#280

### DIFF
--- a/artifacts.py
+++ b/artifacts.py
@@ -31,6 +31,7 @@ def create_openapi_artifacts(openapiart_class):
     ],
         artifact_dir=os.path.join(os.path.dirname(__file__), "art"),
         extension_prefix="sanity",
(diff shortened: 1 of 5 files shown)


ashutshkumr (changes requested):
Overall, the changes look good. Some nits that we should take care of:
- Do not use `/**/` everywhere. Preserve their usage only for file heading. For all others, use `//` (even if the messages are multiline)
- Keep the comment length on each line to be close to 75 chars (max 80). With current test yaml we may not be able to check it, but we should ensure the same with `models` yaml.
- File heading is missing (check the contents of redocly main page to see what file heading should consist of)
- As part of the comments, provide `Default: value` as a separate line (in the very end)
- For required fields, add an annotation `Required: true`

ashutshkumr (approved):

Issue #288, opened 2022-01-10
Unable to set empty string or 0 integer value for required field


jkristia:
It is currently not possible to set a value of "" (empty string) or 0 for an for an integer on a required field.
object.FromJson(json data) returns error

```
func (ctrl *propertiesController) SetPropertyData(w http.ResponseWriter, r *http.Request) {
	var item models.PropertySetDataRequest
	if r.Body != nil {
		body, _ := ioutil.ReadAll(r.Body)
		if body != nil {
			item = models.NewPropertySetDataRequest()
			err := item.FromJson(string(body)) <-- ERROR returned if string is empty or integer is 0
			if err != nil {
				item = nil
			}
		}
	}
```
Work around is to **not** mark the field as required.

ajbalogh:
I disagree with the proposed change. 
Required fields should remain as required as there is a discrepancy between openapi.yaml and the generated .proto and documenting a field as required but marking it as optional only confuses those looking at the .proto file.

The underlying generated fluent wrapper should account for required fields being set/not set, using some internal approach of marking required fields as dirty/not set/touched etc.

PR #290, opened 2022-01-20
Updating Primitive types to Optional fields

ANISH-GOTTAPU:
To address https://github.com/open-traffic-generator/openapiart/issues/288

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -67,14 +67,14 @@ class FluentNew(object):
         self.method_description = None
         self.interface_fields = []
 
-    def isOptional(self, property_name):
-        if self.schema_object is None:
-            return True
-        if "required" not in self.schema_object:
-            return True
-        if property_name not in self.schema_object["required"]:
-            return True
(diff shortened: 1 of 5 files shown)


ajbalogh (commented):
If the field is marked as optional it no longer adheres to the model. Anyone implementing the .proto independently of the non-conforming go wrapper will assume that all fields are optional.

ajbalogh:
I disagree with the proposed change.
Required fields should remain as required as there is a discrepancy between openapi.yaml and the generated .proto and documenting a field as required but marking it as optional only confuses those looking at the .proto file.

The underlying generated fluent wrapper should account for required fields being set/not set, using some internal approach of marking required fields as dirty/not set/touched etc.

ANISH-GOTTAPU:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto

ashutshkumr:
Fixed by https://github.com/open-traffic-generator/openapiart/pull/412

Issue #182, opened 2021-09-21
Add linter to the python and go generated Code in CICD Infra


Rangababu-R:
**Is your feature request related to a use case? Please describe.**
Need a linting validation on CICD for the generated python and go code

**Describe the solution you'd like**
Linting gives more clear picture on the generated code

**Describe alternatives you've considered**

**Additional context**

PR #299, opened 2022-01-28
CI/CD Python Linter

ANISH-GOTTAPU:
CI/CD Python Linter


ANISH-GOTTAPU:
To address https://github.com/open-traffic-generator/openapiart/issues/182

do.py
```diff
@@ -112,19 +116,26 @@ def init():
 
 
 def lint():
-    paths = [pkg()[0], "tests", "setup.py", "do.py"]
+    paths = [
+        pkg()[0],
+        "art/sanity/sanity.py",
```
alakendu, line 121: It looks like we are covering all python files. In that case we can provide root directory

openapiart/common.py
```diff
@@ -12,10 +12,6 @@
 import sanity_pb2_grpc as pb2_grpc
 import sanity_pb2 as pb2
```
alakendu, line 14: ignore common.py file

openapiart/openapiartprotobuf.py
```diff
@@ -167,11 +223,15 @@ def _get_field_type(self, property_name, openapi_object):
...
             if type == "array":
-                item_type = self._get_field_type(property_name, openapi_object["items"])
+                item_type = self._get_field_type(
+                    property_name, openapi_object["items"]
+                )
                 format = openapi_object.get("format")
                 min = openapi_object.get("minimum")
                 max = openapi_object.get("maximum")
-                if (min is not None and min > 2147483647) or (max is not None and max > 2147483647):
+                if (min is not None and min > 2147483647) or (
```
alakendu, line 232: Could you please let me know linter will automatically modify and update in github repository (files within openapiart)/ Need to change locally and push those change.

do.py
ANISH-GOTTAPU: sanity directory has pb2 files which are auto-generated. Touching them would create problem and other files are json and .proto. IMO, generated file is enough to check for linting

openapiart/openapiartprotobuf.py
ANISH-GOTTAPU: black will auto format, but errors from cornflakes we have to change locally and push. If any linting error with cornflakes, CI/CD would fail

openapiart/common.py
ANISH-GOTTAPU: ok, added ignore common.py to flake8

alakendu (approved):

PR #300, opened 2022-01-31
return Byte and hack for warning

alakendu:
This should address https://github.com/open-traffic-generator/snappi/issues/150


Rangababu-R:
@alakendu
from the issues of https://github.com/open-traffic-generator/snappi/issues/150

> 
> issue1: I tried to generate the snappi code and seen the import error, then I realized few lines of code which deletes the generated pb lib files, in setup.py of snappi repo which needs to remove. (a new PR is required in snappi repo as well)
> 
> issue2: could you point me to a unit test. I was unable to review the response with warnings as empty list.

Rangababu-R (approved):

PR #303, opened 2022-02-03
go-versions cicd

Rangababu-R:


ashutshkumr (approved):
Removed need for providing go version when installing dependencies

Issue #304, opened 2022-02-04
x-field-pattern feature auto generates incorrect property


ajbalogh:
**Version information**  
0.1.37

**Describe the bug**  
The openapiart bundler currently generates an incorrect and inadequate openapi component schema property for the extension `x-field-pattern` feature `auto`. 

**To Reproduce**  
Given the following x-field-pattern declaration in a component/schema:

```yaml
Flow.Ipv4:
  type: object
  properties:
        protocol:
          x-field-pattern:
            description: |-
              Protocol, default is 61 any host internal protocol
            format: integer
            length: 8
            default: 61
            features: [auto, count, metric_group]
```

The bundler currently emits the following `auto` property (note there is no default, the type is a placeholder and is incorrect and there is no description):
```yaml
        auto:
          type: string
          enum:
          - auto
          default: auto
```

**Expected behavior**  
The MODELGUIDE.md should be updated to reflect the following:

The auto property indicates to the OTG implementer the following:

- Anywhere there is an "auto" option a default value MUST also be present
- The "auto" option if present in the model, overrides the default value and indicates to the implementer that the OTG should generate a value for the field
- If auto is selected by the user and the system has no alternate value for the field, then the default value MUST be used

Given the `x-field-pattern` declaration in the steps to reproduce the openapiart `bundler` should generate an `auto` property something like the following (needs to include the type, length and default values from the model) :

```yaml
        auto:
          description: |-
            The OTG implementation can provide a system generated 
            value for this property. If the OTG is unable to generate a value 
            the default value must be used.
          type: integer
          default: 61
          minimum: 0
          maximum: 255
```

**Additional context**  
N/A

Rangababu-R:
Available in 0.1.38

PR #306, opened 2022-02-08
Auto Model fix

Rangababu-R:
https://github.com/open-traffic-generator/openapiart/issues/304


ajbalogh (changes requested):
The changes LGTM but please add a specific test for this auto case.
The test should confirm that the following has been generated when auto is present: description, type,  default etc...

Rangababu-R:
> The changes LGTM but please add a specific test for this auto case. The test should confirm that the following has been generated when auto is present: description, type, default etc...

@ajbalogh, Added unit tests that would validate the auto feature.

ajbalogh (approved):

ANISH-GOTTAPU (approved):

PR #314, opened 2022-02-23
Printing Go list Object give invalid output

Rangababu-R:


openapiart/openapiartgo.py
```diff
@@ -2119,11 +2119,14 @@ def _validate_struct(self, new, field):
...
             )
         body += """
-            if obj.obj.{name} != nil {{
+            if {condition} {{
                 {body}
             }}
         """.format(
-            name=field.name, body=inner_body
+            body=inner_body,
+            condition="len(obj.obj.{name}) != 0".format(name=field.name)
```
alakendu, line 2127: is there any case that list object may nil. In that case we need to add both the condition. I mean
```go
 obj.obj.PortMetrics != nil && len(obj.obj.PortMetrics) != 0
```
please ignore if this situation will never arise

Rangababu-R: yeah, this will never arise

alakendu (approved):

PR #321, opened 2022-04-13
Remove Setter for Auto Field

Rangababu-R:


ashutshkumr:
@Rangababu-R could you please add a test for validating auto fields (i.e. we shouldn't have to provide an argument when calling auto)

Rangababu-R:
> @Rangababu-R could you please add a test for validating auto fields (i.e. we shouldn't have to provide an argument when calling auto)

@ashutshkumr  addressed

ashutshkumr:
@Rangababu-R I reviewed the test. I think we need one more test for each python and go where user explicitly sets auto (after having previously set some non-auto choice).

ashutshkumr:
@ajbalogh FYI, we encountered a problem with workflow when working with `auto`.

`auto` is the only attribute so far in snappi, whose setter and getter have same signature (since `obj.Auto()` shouldn't accept any signature).

Hence we thought getting rid of the setter for auto altogether would be a good idea. For choices that are not set to auto, users can call a getter which will automatically change the choice back to auto. But as a cautionary remark, this assumes that auto shall only be used in choices (and might require updating the modelling guide with more info).

Could you please review and and let us know if you see any concern.

ashutshkumr (approved):

PR #324, opened 2022-04-18
Dist Installation Fix

Rangababu-R:
test_requirements is getting populated with "sanity" package too, which shall be ignored


do.py
```diff
@@ -101,30 +102,32 @@ def setup_ext():
...
 
 def setup():
-    run(
-        [
-            py() + " -m pip install --upgrade pip",
-            py() + " -m pip install --upgrade virtualenv",
-            py() + " -m virtualenv .env",
-        ]
-    )
+    if platform.system().lower() == "windows":
```
ashutshkumr, line 105: do we need to check explicitly for windows ? isn't it true for all platforms ?

do.py
```diff
@@ -101,30 +102,32 @@ def setup_ext():
...
+                py() + " -m pip install --upgrade pip",
+                py() + " -m pip install --upgrade virtualenv",
+                py() + " -m virtualenv .env",
+            ]
+        )
 
 
 def init():
     if sys.version_info[0] == 3:
         run([py() + " -m pip install black=={}".format(BLACK_VERSION)])
```
ashutshkumr, line 130: why don't we just move black installation to requirements.txt ?

test_requirements.txt
```diff
@@ -1,5 +1,5 @@
-pytest
-openapiart.egg
+flask
 openapiart
-Flask
-sanity
+pytest
+flake8
+black==22.1.0 ; python_version > '2.7'
```
ashutshkumr, line 5: why should black and openapiart be in test_requirements.txt ?

also when should do.py install test_requirements.txt ?

ashutshkumr (changes requested):
needs some minor changes

do.py
Rangababu-R: yes, virtualenv module is supported only for python2.7 in windows and in python3 venv is build-in

ashutshkumr: I mean we can remove the check for windows right ? venv should be available for all platforms for python3.6+.

Rangababu-R: I doubt for linux

ashutshkumr: https://docs.python.org/3/library/venv.html suggests it should be supported for all platforms.

ashutshkumr:
Following line will need to be removed from openapiart's setup.py
```
installation_requires.append("black==22.1.0 ; python_version > '2.7'")
```

ashutshkumr:
When using `do.py` to install openapiart, it works cause we directly use requirements.txt which has a `prefer-binary` flag.

But if someone tries to install openapiart as a package directly, we get rid of that flag (in setup.py).
I remember we used this flag to allow installation on windows. Does installation work when installing openapiart as a package and not using requirements.txt ?

Rangababu-R:
> When using `do.py` to install openapiart, it works cause we directly use requirements.txt which has a `prefer-binary` flag.
> 
> But if someone tries to install openapiart as a package directly, we get rid of that flag (in setup.py). I remember we used this flag to allow installation on windows. Does installation work when installing openapiart as a package and not using requirements.txt ?

yes it works when installing openapiart as package. when package distribution is generated this flag also gets added to it. I had unzipped the wheel file and cross checked.

Rangababu-R:
> We should upgrade grpcio to 1.44.0 as it seems to work on macbook with M1 chip too.
> 
> ```
> grpcio==1.44.0 ; python_version > '2.7'
> ```

changing the version, giving weird issues in CI/CD
https://github.com/open-traffic-generator/openapiart/actions/runs/2187119373

ashutshkumr (approved):

Issue #328, opened 2022-05-14
Add support for external grpc client


ajbalogh:
**Is your feature request related to a use case? Please describe.**
I'd like to file a FR for gosnappi to support creating a gRPC transport from a connection, currently you can only specify a "location" which is the end address without being able to attach any gRPC metadata or use a custom dialer (useful for going through a proxy) For example adding a method to the GrpcTransport interface (https://github.com/open-traffic-generator/snappi/blob/322667fca511b5f111ed0771bdebb5d28ff136f6/gosnappi/common.go#L22) such as SetClientConn (or similar) that takes a *grpc.ClientConn Is this possible? This blocks our automated traffic testing internally

Opened on behalf of alexmasi@google.com

ashutshkumr:
@alakendu could you please take a first look ?

I'll be more willing to have a method which gives you the underlying transport object, so one can modify it like they would modify a usual grpc conn.

alexmasi:
Here would be my preferred solution, consider the example for an Ondatra binding:

```
func (a *kneATE) DialOTG() (gosnappi.GosnappiApi, error) {
        ctx := context.Background()
	addr := "addr_of_proxy" // my custom proxy address
	opts := []*grpc.DialOptions{ /* my custom options required for the proxy chain */ }
	conn, err := grpc.DialContext(ctx, addr, opts...)
	if err != nil {
		return nil, err
	}

	api := gosnappi.NewApi()
	api.NewGrpcTransport().
		SetClient(otg.NewOpenapiClient(conn))
	return api, nil
}
```

Now the api object already has a non-nil grpcClient field and even the first method call would not need to create a new grpcClient. From my understanding of the flow if SetConfig was the first method call:

```
func (api *gosnappiApi) SetConfig(config Config) (ResponseWarning, error) {

	err := config.Validate()
	if err != nil {
		return nil, err
	}

	if api.hasHttpTransport() {
		return api.httpSetConfig(config)
	}

	if err := api.grpcConnect(); err != nil {  // normally this would dial the location, however if we created the GosnappiApi like above then this would be a no-op
		return nil, err
	}
	request := otg.SetConfigRequest{Config: config.Msg()}
	ctx, cancelFunc := context.WithTimeout(context.Background(), api.grpc.requestTimeout)
	defer cancelFunc()
	resp, err := api.grpcClient.SetConfig(ctx, &request)
	if err != nil {
		return nil, err
	}
	if resp.GetStatusCode_200() != nil {
[...]

PR #329, opened 2022-05-17
Add support for external grpc client

alakendu:
resolves https://github.com/open-traffic-generator/openapiart/issues/328

We will expose these two methods in GrpcTransport interface
```go
	// SetClientConnection set grpc DialContext
	SetClientConnection(con *grpc.ClientConn) GrpcTransport
	// ClientConnection get grpc DialContext
	ClientConnection() *grpc.ClientConn
```
This is one snippet how user call it

```go
	ctx, cancelFunc := context.WithTimeout(context.Background(), 20*time.Second)
	defer cancelFunc()
	conn, err := grpc.DialContext(ctx, grpcServer.Location, grpc.WithTransportCredentials(insecure.NewCredentials()))
	if err != nil {
		log.Fatal(fmt.Sprintf("failed grpc dialcontext due to %s", err.Error()))
	}
	clientConnApi := openapiart.NewApi()
	clientConnApi.NewGrpcTransport().SetClientConnection(conn)
```

### DIFF
--- a/openapiart/common.go
+++ b/openapiart/common.go
@@ -17,15 +17,15 @@ type grpcTransport struct {
 }
 
 type GrpcTransport interface {
-    // SetLocation set client connection to the given grpc target
+	// SetLocation set client connection to the given grpc target
 	SetLocation(value string) GrpcTransport
 	// Location get grpc target
 	Location() string
-    // SetRequestTimeout set timeout in grpc request
+	// SetRequestTimeout set timeout in grpc request
 	SetRequestTimeout(value time.Duration) GrpcTransport
-    // RequestTimeout get timeout in grpc request
+	// RequestTimeout get timeout in grpc request
 	RequestTimeout() time.Duration
-    // SetDialTimeout set timeout in grpc dial
+	// SetDialTimeout set timeout in grpc dial
 	SetDialTimeout(value time.Duration) GrpcTransport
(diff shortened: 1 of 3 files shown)


openapiart/common.go
```diff
@@ -17,12 +17,23 @@ type grpcTransport struct {
 }
 
 type GrpcTransport interface {
+    // SetLocation set client connection to the given grpc target
```
ajbalogh, line 20: minor nit, fix indenting on this line and other comments in this interface

ajbalogh (changes requested):
lgtm except for minor indenting

ajbalogh (approved):

alexmasi:
Anything stopping this from being merged?

ajbalogh:
@alexmasi the change looks good but its missing a unit test which will be addressed asap.

ajbalogh (changes requested):
Missing a unit test.

alakendu:
I have created another api  `clientConnApi`  using SetClientConnection() which will use at the time of other UT execution. 
But I agree it will better to add additional UT. Please check this UT `TestGrpcClientConnection`
https://github.com/open-traffic-generator/openapiart/blob/main/pkg/transport_test.go

ajbalogh (approved):

PR #334, opened 2022-05-24
Http Close API

Rangababu-R:


pkg/transport_test.go
```diff
@@ -166,8 +221,28 @@ func TestConnectionClose(t *testing.T) {
...
+	httpApi.NewHttpTransport().SetLocation(httpServer.Location)
+	config1 := NewFullyPopulatedPrefixConfig(httpApi)
+	config1.SetResponse(openapiart.PrefixConfigResponse.STATUS_200)
+	resp1, err1 := httpApi.SetConfig(config1)
 	assert.Nil(t, err1)
+	assert.NotNil(t, resp1)
+
+	err2 := api.Close()
+	assert.Nil(t, err2)
+	data := NetStat("ESTABLISHED", t)
```
ashutshkumr, line 235: can we avoid checking only for ESTABLISHED state ? a lingering tcp connection may be in non-established state too and still cause issues.

openapiart/openapiartgo.py
```diff
@@ -642,12 +642,32 @@ def _build_api_interface(self):
...
+                            }}
+                            tlsConn := tls.Client(tcpConn, &tls.Config{{InsecureSkipVerify: !api.http.verify}})
+                            err = tlsConn.Handshake()
+                            if err != nil {{
+                                return nil, err
+                            }}
+                            api.http.conn = tcpConn
+                            return tlsConn, nil
+                        }},
+                        DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {{
```
ashutshkumr, line 659: we shouldn't need this right ? let's remove it.

ashutshkumr (changes requested):
need to fix test validating netstat output

openapiart/openapiartgo.py
Rangababu-R: let me check if we remove this will our tests work properly, as the test servers runs on http

Rangababu-R: If removed the UTs are failing. as test server is serving http responses.

pkg/transport_test.go
Rangababu-R: updated

ashutshkumr (approved):

PR #336, opened 2022-05-30
Enhance CI to check installation of generated SDK

ASHNA-AGGARWAL-KEYSIGHT:
resolves https://github.com/open-traffic-generator/openapiart/issues/333


do.py
```diff
@@ -177,13 +176,15 @@ def lint():
     )
 
 
-def generate():
+def generate(sdk="", cicd=""):
     artifacts = os.path.normpath(
         os.path.join(os.path.dirname(__file__), "artifacts.py")
     )
+    if cicd:
+        run([py() + " -m pip install openapiart"])
```
Rangababu-R, line 184: this statement should not be there Ashna. remove it. in cicd the openapiart shall be installed from the git workflow which is already done.

Rangababu-R (changes requested):

Rangababu-R (approved):

PR #337, opened 2022-06-07
Error Messages Consistency

### LINKED ISSUE #221: Make error messages consistent and uniform across python and go SDK

### LINKED ISSUE #274: Enum Setter and Getter behaviour is incorrect for invalid choices

### LINKED ISSUE #276: Object type, required property is not being validated

### LINKED ISSUE #277: Validation errors are not clubbed

Rangababu-R:
resolves #221 
resolves #277 
resolves #276 
resolves #274


ashutshkumr:
Covered by https://github.com/open-traffic-generator/openapiart/pull/449

PR #338, opened 2022-06-07
Support for gosnappi Clone method

### LINKED ISSUE #273: Object cloning method is unavailabe in gosnappi

Rangababu-R:
resolves #273

### DIFF
--- a/openapiart/openapiartgo.py
+++ b/openapiart/openapiartgo.py
@@ -1086,12 +1086,16 @@ class OpenApiArtGo(OpenApiArtPlugin):
             }}
 
             func (obj *{struct}) Clone() ({interface}, error) {{
+                vErr := obj.Validate()
+                if vErr != nil {{
+                    return nil, vErr
+                }}
                 newObj := New{interface}()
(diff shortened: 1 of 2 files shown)


openapiart/openapiartgo.py
```diff
@@ -1063,6 +1063,19 @@ def _write_interface(self, new):
                 }}
                 return str
             }}
+
+            func (obj *{struct}) Clone() ({interface}, error) {{
+                newObj := New{interface}()
+                pbText, err := obj.ToPbText()
```
ashutshkumr, line 1069: I would recommend marshelling to proto bytes instead of proto text (since it will be more efficient).

https://gist.github.com/miguelmota/25568433ad8cfddb5ea556a5644c9fde#file-main-go-L16

ashutshkumr (changes requested):
please sync this branch with main as well

ashutshkumr (approved):

PR #340, opened 2022-06-07
Support for deprecation message of a property

### LINKED ISSUE #319: implement x-status constraint
x-status constraint is already available in models which will provide the validity of a schema object or a property. Implementation needs to be addressed

To provide the user the schema object or property is valid, deprecated, obsolete or under review

Rangababu-R:
resolves #319


Rangababu-R:
addressed in #354

Issue #326, opened 2022-04-18
SDK must validate uniqueness of an object name


ashutshkumr:
**Is your feature request related to a use case? Please describe.**
snappi should be able to validate uniqueness of object name within a certain scope. Dependency on Related to https://github.com/open-traffic-generator/models/issues/233

Issue #325, opened 2022-04-18
Implement x-constraint for names


ashutshkumr:
**Is your feature request related to a use case? Please describe.**
As a user, I should be able to rely on python or Go SDK to raise validation error when violating x-constraint for object names as described in models. e.g. A field that accepts port names is set with a device name

Issue #195, opened 2021-10-05
Add new format keyword global-unique


ajbalogh:
### Is your feature request related to a use case? Please describe.
- Users currently have to add a unique name for every object that has a property documented as being globally unique.
  ```yaml
  properties:
    name:
      description: >-
        Globally unique name of an object.
        It also serves as the primary key for arrays of objects.
      type: string
      pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
  ```
  ```python
  p1, p2 = config.ports.port(name='p1').port(name='p2')
  
  c1 = config.capture.capture(name='c1')
  c1.port_names = [p1.name]
  ```
 
 ### Describe the solution you'd like
```python
# if the user does not specify a name,  one will be generated for them
p1, p2 = config.ports.port().port()
print(p1, p2) # outputs PORT-1 PORT-2
```
- Instead of implicit descriptions there should be an explicit string format keyword named `global-unique` that is documented in the modeling guide
- The keyword MUST be supported by the python/go ux sdk generators which will generate a unique name if none is provided.
- The python/go ux sdk generators should generate a named based on the schema object that contains the property, the format of the name MUST be `<schema object in uppercase>-<next available int>`.
  ```yaml
  # python/go sdk generation should produce a default name <schema object name in upper case>-<next available int>
  # i.e., name: PORT-1
  Port:
    type: object
    properties:
      name:
      type: string
      format: global-unique
      pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$    
  ```
- Validation that is part of the python/go ux sdk packages MUST enforce the keyword by ensuring that all properties with a format of `global-unique` are globally unique.
  ```yaml
  properties:
    name:
      description: >-
        Globally unique name of an object.
        It also serves as the primary key for arrays of objects.
      type: string
      format: global-unique
      pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
  ```

### Describe alternatives you've considered
The current OpenApi specificiation does not provide any format for indicating a globally unique name.

### Additional context
N/A

PR #345, opened 2022-07-01
Unique and Constraint Validation

Rangababu-R:
Proposing x-unique field to the models to validate the unique properties of an objects.
```
components:
  schemas:
    Prefix.Config:
      properties:
        w_list:
          type: array
          items:
            $ref: '#/components/schemas/WObject'
        x_list:
          type: array
          items:
            $ref: '#/components/schemas/ZObject'
        name:
          type: string
          x-unique: global

    ZObject:
      required: [name]
      properties:
        name:
          type: string
          x-unique: local

    WObject:
      required: [w_name]
      properties:
        w_name:
          type: string
          x-unique: global  
```
An object which shall be unique only for local can be validated against the list of objects of its own type

resolves https://github.com/open-traffic-generator/openapiart/issues/326
resolves https://github.com/open-traffic-generator/openapiart/issues/325
resolves https://github.com/open-traffic-generator/openapiart/issues/195


Rangababu-R:
@ajbalogh could you please provide your review on this PR

PR #354, opened 2022-07-26
X unique, X constraint and deprecate

Rangababu-R:


ashutshkumr:
Follow language-specific idioms to deprecate things. e.g. in Go, following needs to be the format (per https://stackoverflow.com/questions/48764572/how-to-correctly-deprecate-api-endpoint-in-go-goa-framework)
```go
// Deprecated: Use strings.HasPrefix instead.
```

Similarly in python https://stackoverflow.com/questions/30093490/how-to-declare-a-module-deprecated-in-python
- Note, the warning will need to be raised from inside func and not at module level (please feel free to look for better alternative)

ashutshkumr:
Is it possible to avoid storing warnings/errors globally ? The issue is if a user is using package to work with two different objects, we'll possibly end up concatenating warnings/errors among one another.

```go
var validation []string
var unique_global []string
var constraints = make(map[string]map[string]interface{})
var openapi_warnings []string
```

We should add a TC to cover this scenario.
Moreover, I'm curious what happens if we make multiple API calls each of which raises different warning.

openapiart/common.go
```diff
@@ -305,3 +335,85 @@ func validateIpv6Slice(ip []string) error {
...
+		create(objectName)
+		constraints[objectName][value] = object
+		return !ok
+	}
+	_, ok = values[value]
+	if !ok {
+		constraints[objectName][value] = object
+		return !ok
+	}
+	return !ok
```
ashutshkumr, line 374: this portion will never be executed, so can we please remove it ?

ashutshkumr (approved):

PR #356, opened 2022-08-17
Isolate Bundling

Rangababu-R:
This is to Isolate bundling from constructor and have control over it.

### DIFF
--- a/artifacts.py
+++ b/artifacts.py
@@ -44,14 +44,20 @@ def create_openapi_artifacts(openapiart_class, sdk=None):
         artifact_dir=os.path.join(os.path.dirname(__file__), "art"),
         extension_prefix="sanity",
         proto_service="Openapi",
+        skip_bundle=True
     )
+    openapi_yaml = open_api.bundle()
     if sdk == "python" or sdk is None:
-        open_api.GeneratePythonSdk(package_name="sanity")
+        open_api.GeneratePythonSdk(
+            package_name="sanity",
+            openapi_yaml=openapi_yaml,
+        )
 
     if sdk == "go" or sdk is None:
         open_api.GenerateGoSdk(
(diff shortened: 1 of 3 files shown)


ajbalogh (approved):

alakendu:
looks like `openapi_yaml` for `GeneratePythonSdk` and `GenerateGoSdk` are disjoin and require knowledge that pass return of bundler.
I am thinking what about if we pass `bundled_file` within constructor which is `None` by default. 
```python
open_api = openapiart.OpenApiArt(
    api_files=["openapi.yaml"],
    bundled_file = None
)
```
We will raise error if user provide `bundled_file` as well as called `open_api.bundle()`

Rangababu-R:
> looks like `openapi_yaml` for `GeneratePythonSdk` and `GenerateGoSdk` are disjoin and require knowledge that pass return of bundler. I am thinking what about if we pass `bundled_file` within constructor which is `None` by default.
> 
> ```python
> open_api = openapiart.OpenApiArt(
>     api_files=["openapi.yaml"],
>     bundled_file = None
> )
> ```
> 
> We will raise error if user provide `bundled_file` as well as called `open_api.bundle()`

Actually the purpose of bundle isolation is because of "enum" and "x-enum" can't co-live. prior to this bundling with openapi.yaml also works fine.
So, skip_bundle will provide the linkage between bundling and as well as the sdk generation. the only thing we have to add is once this PR is merged we need to raise the exception if "enum" and "x-enum" co-lives then the message shall be "enum and x-enum shall not co-live and do not try to bundle and already bundled openapi.yaml"
This is enough, I feel bundled_file param is extra when we are isolating the bundle that is why I did not add bundled_file param (which initially I thought of to do so)

ashutshkumr:
Stale PR

Issue #358, opened 2022-08-26
Need support for marshalling and unmarshalling proto object with defaults


ashutshkumr:
**Is your feature request related to a use case? Please describe.**
When implementing gRPC server, we receive request as a proto object without guarantees of defaults having been set.
Hence a helper is needed to take in proto object and set defaults - similar to `FromJson`.

Without this, handler has to do redundant marshalling/unmarshalling for setting defaults.

**Describe the solution you'd like**
```go
in := &otg.GetMetricsRequest{}
out := gosnappi.NewApi().NewMetricsRequest()

// will internally set proto object and set defaults
if err := out.FromProto(in); err != nil {
  return err
}

// should output proto object with defaults set (unlike out.Msg())
in, err := out.ToProto()
err != nil {
  return err
}
```

**Describe alternatives you've considered**
```go
in := &otg.GetMetricsRequest{}
out := gosnappi.NewApi().NewMetricsRequest()

// this won't set the default values
out.SetMsg(in)

// hence serialize to pb text (this also won't set the default values)
pb, err := rbody.ToPbText()
if err != nil {
    return err
}
// deserialize from pb text (this will set the default values)
if err := out.FromPbText(pb); err != nil {
  return err
}
```

ashutshkumr:
@ajbalogh kindly review the proposal

ashutshkumr:
Fixed in latest

PR #361, opened 2022-08-29
To And From Proto

Rangababu-R:
#358


ajbalogh (approved):
LGTM but should there be a doc string to indicated the difference between Msg/SetMsg since the signatures are the same?

Rangababu-R:
> LGTM but should there be a doc string to indicate the difference between Msg/SetMsg since the signatures are the same?

addressed

PR #366, opened 2022-09-20
X-Constraint List value validation fix

Rangababu-R:
https://github.com/open-traffic-generator/openapiart/issues/367

### DIFF
--- a/openapiart/common.go
+++ b/openapiart/common.go
@@ -128,7 +128,7 @@ type Api interface {
 	// and not intended to use in production
 	Warnings() string
 	deprecated(message string)
-	under_review(message string)
+	underReview(message string)
 }
 
 // NewGrpcTransport sets the underlying transport of the Api as grpc
@@ -175,7 +175,7 @@ func (api *api) deprecated(message string) {
 	fmt.Printf("warning: %s\n", message)
 }
 
-func (api *api) under_review(message string) {
+func (api *api) underReview(message string) {
 	api.warnings = message
 	fmt.Printf("warning: %s\n", message)
 }
@@ -192,46 +192,46 @@ type httpClient struct {
 
 // All methods that perform validation will add errors here
 // All api rpcs MUST call Validate
-type Constraints interface {
-	ValueOf(name string) interface{}
+type valueGetter interface {
+	StringValue(name string) (string, error)
 }
 
-type rootObject interface {
-	validateObj(set_default bool)
+type rootType interface {
+	validateObj(setDefault bool)
 	validationResult() error
 	checkUnique()
 	checkConstraint()
 }
 
-type validation struct {
-	validationErrors []string
-	warnings         []string
-	constraints      map[string]map[string]Constraints
-	rootObj          rootObject
-	temp             []rootObject
-	resolve          bool
+type validator struct {
+	errors      []string
+	warnings    []string
+	constraints map[string]map[string]valueGetter
+	root        rootType
+	temp        []rootType
+	resolve     bool
 }
 
-type Validation interface {
+type Validator interface {
 	validationResult() error
 	deprecated(message string)
-	under_review(message string)
+	underReview(message string)
 	Warnings() []string
 }
 
-func (obj *validation) validationResult() error {
-	obj.constraints = make(map[string]map[string]Constraints)
+func (obj *validator) validationResult() error {
+	obj.constraints = make(map[string]map[string]valueGetter)
 	obj.temp = nil
-	if len(obj.validationErrors) > 0 {
-		obj.validationErrors = append(obj.validationErrors, "validation errors")
-		errors := strings.Join(obj.validationErrors, "\n")
-		obj.validationErrors = nil
+	if len(obj.errors) > 0 {
+		obj.errors = append(obj.errors, "validation errors")
+		errors := strings.Join(obj.errors, "\n")
+		obj.errors = nil
 		return fmt.Errorf(errors)
 	}
 	return nil
 }
 
-func (obj *validation) Warnings() []string {
+func (obj *validator) Warnings() []string {
 	if len(obj.warnings) > 0 {
 		warns := obj.warnings
 		obj.warnings = nil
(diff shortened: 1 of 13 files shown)


ashutshkumr:
Why do we store warnings  as a `string` and not `[]string` ? When we find something deprecated / under-review, we could just append it (instead of re-assigning) right ?
```go
func (api *api) deprecated(message string) {
	api.warnings = message
	fmt.Printf("warning: %s\n", message)
}
```

We might benefit from having `AddWarning()` or `AppendWarning()` method, which will internally log the warning lines when they're being added.

ashutshkumr:
Following line should be moved to its own function, maybe `LogWarning`.
```go
fmt.Printf("warning: %s\n", message)
```

ashutshkumr:
Why don't we implement `Close` on `type api struct` ? I noticed out of 8 methods, only 7 are implemented. That implies `api` doesn't completely implement `Api` interface.

ashutshkumr:
Can we do something about `sanity` naming ? Curious why the proto files are not generated with name `openapiartpb` (derived from the package name itself).

I'm mentioning this because some users raised concerns about the naming of these files itself (they blindly copied the contents of artifact.py).

Another concern, why openapiart_class takes so many YAMLs as input ? Isn't there a single entry point ?

ashutshkumr:
I think it's bad UX to have a type `<package>.<name>Api` instead of `<package>.Api` for objects returned from `<package>.NewApi()`

There's really no incentive for users to use base API class. So I would propose we rename base api class to `BaseApi` or something and rename actual API class to `Api`.

This might cause breaking change, and hence for some time, we might want to keep an alias.

```go
// a should be of type openapiart.Api
a := openapiart.NewApi()

// if a func accepts and object of type openapiart.OpenapiArt, it should also be able to accept object of type openapiart.Api (i.e. one should be an alias of another)
```

ashutshkumr:
There are lots of identifiers with underscore in their name (e.g. under_review, etc). Would suggest renaming those to use camelCase convention (which is idiomatic in Go).

ashutshkumr:
Would suggest renaming following

```go
type validation struct {
	validationErrors []string
	warnings         []string
	constraints      map[string]map[string]Constraints
	rootObj          rootObject
	temp             []rootObject
	resolve          bool
}
```

to 

```go
type typeMeta struct {
	errors              []string
	warnings         []string
	constraints      map[string]map[string]Constraints
	rootType         rootType
        // what is temp for ?
	temp             []rootObject
       // resolve what ?
	resolve          bool
}
```

In general, would like to replace the term `validation` wherever it's more appropriate to call it `typeMeta` and replace `object` with `type`.

ashutshkumr:
`Has*` methods have incorrect documentation.

```go
// Space1 returns a int32
// description is TBD
func (obj *prefixConfig) HasSpace1() bool {
	return obj.obj.Space_1 != nil
}
```

ashutshkumr:
We do not print names of attributes which are under review. e.g. in this case `A`.
Also, curious why we see duplicate prints (is it because of usage of Set* and ToJson ?).

```go
func TestPrefixConfigCustom(t *testing.T) {
	api := openapiart.NewApi()
	c1 := api.NewPrefixConfig()
	c1.RequiredObject()
	c1.SetA("hello")
	c1.SetB(5)

	c, e := c1.ToJson()
	fmt.Println("errors:", e)
	fmt.Println("config:", c)
}
```

Output
```
2022/10/10 10:14:21 MockGrpcServer: Server started and listening on address [::]:40051
2022/10/10 10:14:21 MockGrpcServer: Server subscribed with gRPC Protocol Service
2022/10/10 10:14:21 Generated Http Server serving incoming HTTP requests on  127.0.0.1:50051
=== RUN   TestPrefixConfigCustom
warning: property under review
warning: Property b is being deprecated from the sdk version x.x.x
and property x shall be used instead
warning: property under review
warning: Property b is being deprecated from the sdk version x.x.x
and property x shall be used instead
errors: <nil>
config: {
  "required_object": {},
  "response": "status_200",
  "a": "hello",
  "b": 5,
  "h": true
}
```

ashutshkumr:
openapiart should raise an error if `default` is provided for a required field.

ashutshkumr:
I see lots of commented out code before return statement, can we please get rid of those ?
e.g.

```go
func (api *openapiartApi) NewGetConfigResponse() GetConfigResponse {
	vObj := validation{}
	obj := newGetConfigResponse(&vObj)
	vObj.rootObj = obj
	vObj.resolve = true
	return obj
	// return NewGetConfigResponse()
}
```

Rangababu-R:
> Why do we store warnings as a `string` and not `[]string` ? When we find something deprecated / under-review, we could just append it (instead of re-assigning) right ?
> 
> ```go
> func (api *api) deprecated(message string) {
> 	api.warnings = message
> 	fmt.Printf("warning: %s\n", message)
> }
> ```
> 
> We might benefit from having `AddWarning()` or `AppendWarning()` method, which will internally log the warning lines when they're being added.

we had discussed this before and concluded to print the warning

Rangababu-R:
> Why don't we implement `Close` on `type api struct` ? I noticed out of 8 methods, only 7 are implemented. That implies `api` doesn't completely implement `Api` interface.
> 
> It's probably ok to have a dummy implementation or an implementation that just prints warning that `Close` is not implemented.

please create an issue will be addressed in separate PR

Rangababu-R:
> Can we do something about `sanity` naming ? Curious why the proto files are not generated with name `openapiartpb` (derived from the package name itself).
> 
> I'm mentioning this because some users raised concerns about the naming of these files itself (they blindly copied the contents of artifact.py).
> 
> Another concern, why openapiart_class takes so many YAMLs as input ? Isn't there a single entry point ?

this is not related to x-constraints, unique and deprecate, please create an issue will be addressed in separate PR

Rangababu-R:
> I think it's bad UX to have a type `<package>.<name>Api` instead of `<package>.Api` for objects returned from `<package>.NewApi()`
> 
> There's really no incentive for users to use base API class. So I would propose we rename base api class to `BaseApi` or something and rename actual API class to `Api`.
> 
> This might cause breaking change, and hence for some time, we might want to keep an alias.
> 
> ```go
> // a should be of type openapiart.Api
> a := openapiart.NewApi()
> 
> // if a func accepts and object of type openapiart.OpenapiArt, it should also be able to accept object of type openapiart.Api (i.e. one should be an alias of another)
> ```

this is not related to x-constraints, unique and deprecate, please create an issue will be addressed in separate PR

Rangababu-R:
> We should rename following for more consistency:
> 
>     * `*Holder` -> `*Wrapper`
> 
>     * `Msg()` -> `Proto()`
> 
>     * `SetMsg()` -> `SetProto()`
> 
>     * `obj` -> `proto`
> 
>     * `func (obj *prefixConfig) FromYaml` -> `func (o *prefixConfig) FromYaml` -  since it's idomatic to use single letter in method definition

this is not related to x-constraints, unique and deprecate, please create an issue will be addressed in separate PR

Rangababu-R:
> `Has*` methods have incorrect documentation.
> 
> ```go
> // Space1 returns a int32
> // description is TBD
> func (obj *prefixConfig) HasSpace1() bool {
> 	return obj.obj.Space_1 != nil
> }
> ```
> 
> Moreover, `a int32` is grammatically incorrect.

this is not related to x-constraints, unique and deprecate, please create an issue will be addressed in separate PR

Rangababu-R:
> openapiart should raise an error if `default` is provided for a required field.

this is not related to x-constraints, unique and deprecate, please create an issue will be addressed in separate PR

ashutshkumr:
Latest changes covered by https://github.com/open-traffic-generator/openapiart/pull/453

PR #386, opened 2023-01-13
Segregate proto generation

### LINKED ISSUE #385: Generation of SDK should not be needed for generating the proto file
currently to get the proto file for a yml we need to call go or python generation

just calling the open api art should also generate the proto file

Vibaswan:
Fixes #385

### DIFF
--- a/openapiart/openapiart.py
+++ b/openapiart/openapiart.py
@@ -61,7 +61,6 @@ class OpenApiArt(object):
         self._get_info()
         self._get_license()
         self._document()
-        self._generate_proto_file()
 
     def _get_license(self):
         license_name = self._bundler._content["info"]["license"]["name"]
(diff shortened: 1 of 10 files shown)


Vibaswan:
For some reason not sure why Ranga actually made the change which i am reverting right now
https://github.com/open-traffic-generator/openapiart/commit/8fc7b8d1e559dcf9f759cb67bf277fd4531ef065#diff-c4e7b5e541655c6a631951b4e8fe368a052b0132fe605d13fbad12689fb0c416

But I don't see any problem in keeping it separate.

Vibaswan:
Made changes in CI so that generated proto, yml and json is validated

openapiart/openapiart.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         self._get_info()
         self._get_license()
         self._document()
+        self._generate_proto_file()
```
ashutshkumr, line 64: We should not generate proto by default (as some environments may not have such requirements).

I recommend exposing a func for generating protobuf separately.
```python
art.GenerateProtoDef(
    package_name="sanity"
)
```

Moreover, I would recommend performing `_generate_proto_file` durin both python and Go SDK (if it hasn't already been performed)

ashutshkumr (changes requested):
Please also sync with main

ashutshkumr (approved):

PR #390, opened 2023-01-27
Adding fix for upper case not supported check for property names

### LINKED ISSUE #294: Using properties with camel case quietly fails
**Version information**
0.1.24

**Describe the bug**
Using camel case for a property quietly doesn't work.
Instead of producing an error when generating python code, it fails when using grpc APIs (e.g. ParseDict) to marshal created objects into a GRPC message.

```
components:
  schemas:
    Thing:
      type: object
      properties:
        thingProperty:
```

Note that it works correctly when using `thing_property`.

**Expected behavior**
If camel case is not supported, a clear error message should be generated when running the openapiart tools in a non-confirming openapi definition.
Or alternatively, camel case should be supported.

dipendughosh:
Fixes #294

### DIFF
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -323,16 +323,16 @@ class Bundler(object):
                 self._validate_names("^[+a-zA-Z0-9_]+$", "responses", value)
                 self._check_nested_components(value)
         self._resolve_refs(base_dir, yobject)
-    
+
     def _check_upper_case(self, value):
         upper_case = False
-                
+
         for c in value:
             if c.isupper():
                 upper_case = True
-        
+
         return upper_case
-        
+
     def _validate_names(self, regex, components_key, components):
         if components_key not in components:
             return
@@ -347,13 +347,14 @@ class Bundler(object):
                         )
                     upper_case = self._check_upper_case(name)
                     if upper_case:
-                        raise NameError("*** Property name '{}' is invalid. Only lower case with underscore separated is allowed.***".format(
-                                    value
-                                )
+                        raise NameError(
+                            "*** Property name '{}' is invalid. Only lower case with underscore separated is allowed.***".format(
+                                value
                             )
-        
+                        )
+
             self._content["components"][components_key][key] = value
-        
+
     def _check_nested_components(self, components):
         objects = components["schemas"]
         errors = []
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -348,7 +348,7 @@ class Bundler(object):
                     upper_case = self._check_upper_case(name)
                     if upper_case:
                         raise NameError(
-                            "*** Property name '{}' is invalid. Only lower case with underscore separated is allowed.***".format(
+                            "*** Property name '{}' is invalid. Only lower case letters separated with an underscore is allowed.***".format(
                                 value
                             )
                         )
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -325,7 +325,7 @@ class Bundler(object):
         self._resolve_refs(base_dir, yobject)
 
     def _check_upper_case(self, value):
-        
+
         for c in value:
             if c.isupper():
                 return True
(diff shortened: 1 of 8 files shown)


openapiart/tests/field_uid/property_name.yaml
```diff
@@ -0,0 +1,64 @@
...
+      properties:
+        camelCase:
+          type: boolean
+          x-field-uid: 1
+        PascalCase:
+          type: boolean
+          x-field-uid: 2
+        Upper_case:
+          type: boolean
+          x-field-uid: 3
```
ashutshkumr, line 64: Is the test actually exercising all three use cases ?

openapiart/tests/field_uid/property_name.yaml
```diff
@@ -0,0 +1,64 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  errors:
+                    type: array
+                    items:
+                      type: string
+                    x-field-uid: 1
```
ashutshkumr, line 48: these are not really needed i guess

ashutshkumr: just a simple 200 without any payload should be fine

openapiart/bundler.py
```diff
@@ -336,8 +345,15 @@ def _validate_names(self, regex, components_key, components):
                             "%s property name `%s` contains invalid characters"
                             % (key, name)
                         )
+                    upper_case = self._check_upper_case(name)
+                    if upper_case:
+                        raise NameError("*** Property name '{}' is invalid. Only lower case with underscore separated is allowed.***".format(
```
ashutshkumr: Only lower case with underscore separated is allowed -> Only lower case letters separated with an underscore is allowed

openapiart/bundler.py
```diff
@@ -336,8 +345,15 @@ def _validate_names(self, regex, components_key, components):
                             "%s property name `%s` contains invalid characters"
                             % (key, name)
                         )
+                    upper_case = self._check_upper_case(name)
+                    if upper_case:
```
ashutshkumr, line 349: would suggest directly using calling the method here if the value is not needed anywhere else

openapiart/bundler.py
```diff
@@ -323,7 +323,16 @@ def _process_yaml_object(self, base_dir, yobject):
...
-
+    
+    def _check_upper_case(self, value):
+        upper_case = False
+                
+        for c in value:
+            if c.isupper():
+                upper_case = True
+        
+        return upper_case
```
ashutshkumr, line 334: would suggest breaking early in the function; probably storing False/True in upper_case var is not required

ashutshkumr (changes requested):
@dipendughosh @Vibaswan kindly check why CI is failing

openapiart/tests/field_uid/property_name.yaml
dipendughosh: Only the 1st one will be tested but the remaining are for future users to understand what we are not supporting.

Vibaswan:
@ashutshkumr 
The CI was failing due a error from flake8
I also faced similar issue and suggested @dipendughosh with the fix I used earlier 
https://github.com/open-traffic-generator/openapiart/commit/f461cf37ac88ebed82d814d091183be030ca4d9f

I guess this fix would also be needed 
https://github.com/open-traffic-generator/openapiart/commit/aef08dfbbed1b9c6d80e83b159cb5878b36cfbe0

openapiart/bundler.py
dipendughosh: done

openapiart/bundler.py
dipendughosh: done

openapiart/bundler.py
ashutshkumr: Can't see the change here

openapiart/tests/field_uid/property_name.yaml
dipendughosh: done

openapiart/bundler.py
dipendughosh: done

ashutshkumr (approved):

PR #392, opened 2023-01-28
New refactor nomencleture change

### LINKED ISSUE #391: go generated SDK should have proper nomenclature
Currently the go SDK has all receivers proto and objects named as obj which might lead to difficulty in reading the file, proper nomenclature should be in place

Vibaswan:
Fixes #391 
* renamed function receivers from obj to rx
* renamed proto pointers from obj to pb
* renamed snappi generated struct pointers to ob

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,8 +129,39 @@ jobs:
         run: |
           python do.py testgo
 
+  proto_yml_generation_test:
+    needs: build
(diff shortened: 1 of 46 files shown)


Vibaswan:
models successful run 
https://github.com/open-traffic-generator/models/actions/runs/4220581615/jobs/7326990841

snappi successful run
https://github.com/open-traffic-generator/snappi/actions/runs/4220637340

PR #399, opened 2023-02-15
New set() API support for properties with native type and required property cannot be set with None

### LINKED ISSUE #168: Allow setting attributes with built-in types using a `set()` method in python SDK
The method name is not final. Here's the requirement as specified by Matt B.

Currently we have to do:

```python
def setup_devices_long():
    dev = config.devices.add(name='dev', container_name=config.ports[0].name)
    dev.ethernet.name = 'why_eth'
    dev.ethernet.mac = '00:00:00:00:00:00'
    dev.ethernet.ipv4.name = 'why_ipv4'
    dev.ethernet.ipv4.address = '1.1.1.1'
    dev.ethernet.ipv4.prefix = 20
    dev.ethernet.ipv4.gateway = '1.1.1.254'
```

Proposed way:

```python
def setup_devices():
    dev = config.devices.add(name='dev', container_name=config.ports[0].name)
    dev.ethernet.set(name='why_eth', mac='00:00:00:00:00:00')
    dev.ethernet.ipv4.set(name='why_ipv4', address='1.1.1.1', prefix=20, gateway='1.1.1.254')
```

Few more considerations on impl:
- only values of built-in or [built-in] types should be allowed to be passed as arguments
- function signature should be explicit kwargs with defaults set, for better intellisense (instead of **kwargs/monkey-patching)

### LINKED ISSUE #395: Required properties cannot be set with 'None'.
While implementing a new 'set' api to set all properties at a time it is seen that for required properties there is not check for 'None' values getting set. This needs to be supported else users might set a required property with a 'None' value.
The issue is [Allow setting attributes with built-in types using a set() method in python SDK#168](https://github.com/open-traffic-generator/openapiart/issues/168) which pointed out the current issue.

dipendughosh:
Fixes  #168 - Allow setting attributes with built-in types using a set() method in python SDK.
Fixes #395 - Required properties cannot be set with 'None'.

### DIFF
--- a/openapiart/generator.py
+++ b/openapiart/generator.py
@@ -905,6 +905,35 @@ class Generator:
                 self._write(2, "else:")
                 self._write(3, "self.choice = choice")
 
+            # write def set(self)
+            write_set = False
+            params = "self"
+            if "choice" in self._get_choice_names(schema_object):
+                write_set = False
+            init_params, properties, _ = self._get_property_param_string(
+                schema_object
+            )
+            if len(init_params) > 0:
+                write_set = True
+            params = (
+                params
+                if len(init_params) == 0
+                else ", ".join([params, init_params])
+            )
+
+            if write_set:
+                self._write(1, "def set(%s):" % (params))
+                self._write(
+                    2, "for property_name, property_value in locals().items():"
+                )
+                self._write(
+                    3,
+                    "if property_name != 'self' and property_value is not None:",
+                )
+                self._write(
+                    4, "self._set_property(property_name, property_value)"
+                )
+
             # process properties - TBD use this one level up to process
             # schema, in requestBody, Response and also
             refs = self._process_properties(
@@ -1400,6 +1429,15 @@ class Generator:
             self._write()
             self._write(2, "value: %s" % restriction)
             self._write(2, '"""')
+            required, defaults = self._get_required_and_defaults(schema_object)
+            if len(required) > 0:
+                if name in required:
+                    self._write(2, "if value is None:")
+                    self._write(
+                        3,
+                        "raise TypeError('Cannot set required property %s as None')"
+                        % name,
+                    )
             if write_set_choice is True:
                 self._write(
                     2, "self._set_property('%s', value, '%s')" % (name, name)
--- a/openapiart/generator.py
+++ b/openapiart/generator.py
@@ -907,7 +907,6 @@ class Generator:
 
             # write def set(self)
             write_set = False
-            params = "self"
             if "choice" in self._get_choice_names(schema_object):
                 write_set = False
             init_params, properties, _ = self._get_property_param_string(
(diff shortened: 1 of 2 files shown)


openapiart/generator.py
```diff
@@ -905,6 +905,33 @@ def _write_openapi_object(self, ref, choice_method_name=None):
                 self._write(2, "else:")
                 self._write(3, "self.choice = choice")
 
+            # write def set(self)
```
Vibaswan, line 908: I guess we should have a method which would write the set method `_write_set_method`.
It would help in maintainability and doing further modifications.
for example use have method to write `add` and `factory`:
```python
def _write_add_method:

def _write_factory_method:
```

openapiart/tests/test_set.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    # Use the new set API to set pattern property values of each class
+    config.mac_pattern.mac.increment.set(
+        start="aa:aa:bb:bb:cc:cc", step="00:01:00:00:10:00", count=10
+    )
+    config.integer_pattern.integer.decrement.set(start=200, step=2, count=100)
+
+    config_data = config._properties
+
+    if not config_data["ieee_802_1qbb"]:
```
Vibaswan, line 42: I guess instead of a `if` check it would better we call `assert` for verifying the parameters.
I see rest of the unit test cases use assert and it will good to follow a common approach for the whole test directory
for example
```python
assert config_data["ieee_802_1qbb"] == True, "ieee_802_1qbb value is not set correctly"
``` 
for more reference you can follow the link 
https://www.geeksforgeeks.org/python-assert-keyword/

Vibaswan (changes requested):

openapiart/generator.py
ashutshkumr: @dipendughosh agreed, let's refactor this out to a different method

openapiart/tests/test_set.py
ashutshkumr: yup, assert is more idiomatic for tests

openapiart/generator.py
dipendughosh: Created a new _write_get_method

openapiart/tests/test_set.py
dipendughosh: changed if to assert

openapiart/tests/test_set.py
```diff
@@ -0,0 +1,117 @@
+import sys
+import os
+import pytest
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "..", "art"))
+import sanity
+
```
ashutshkumr, line 7: I'll suggest couple of things here:
- we should serialize at the end to make sure we don't see any exceptions
- we should add some negative cases, i.e. try setting some invalid value or type, it should throw exception (either during set or serialize)

Vibaswan (approved):

ashutshkumr (approved):

openapiart/tests/test_set.py
dipendughosh: submitted the changes is test

ashutshkumr (approved):

PR #400, opened 2023-02-15
Allow choices with no properties

### LINKED ISSUE #365: Allow choices with no properties
Allow choices with no properties.

Don't require a body object for choices with no properties. Choices with at least one selection not having any properties is a valid use case. For example, pick a filter. No filter is a valid choice and would not have any parameters where pretty much every other type of filter would have parameters. 

**Describe alternatives you've considered**
Building unnatural models, like referring to such choices by name.

**Additional context**

Vibaswan:
Fixes #365
changes to support choices with no properties
also added test cases for the same

### DIFF
--- a/openapiart/tests/test_choice_with_no_property.py
+++ b/openapiart/tests/test_choice_with_no_property.py
@@ -90,5 +90,32 @@ def test_choice_in_choice_heirarchy(api):
     assert c_obj.f_obj.choice == "f_c"
 
 
+def test_choice_with_invalid_enum_and_none_value(api):
+    config = api.prefix_config()
+    f_obj = config.f
+
+    # check default
+    assert f_obj.choice == "f_a"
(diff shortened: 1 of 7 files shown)


openapiart/tests/test_choice_with_no_property.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # serialize and deserialize should have no problem
+    s_c_obj = c_obj.serialize()
+    c_obj.deserialize(s_c_obj)
+    assert c_obj.choice == "f_obj"
+    assert c_obj._properties.get("f_obj") is not None
+    assert c_obj.f_obj.choice == "f_c"
+
+
+def test_choice_with_invalid_enum_and_none_value(api):
```
Vibaswan, line 93: added no test for setting choice valid , invalid and None

ashutshkumr (approved):

PR #402, opened 2023-02-22
X status warnings for go and python code

### LINKED ISSUE #319: implement x-status constraint

Vibaswan:
Fixes #319
added x-status support for go and python SDK, added tests as well

### DIFF
--- a/do.py
+++ b/do.py
@@ -276,7 +276,7 @@ def go_lint():
     )
     run([pkg])
     os.chdir("pkg")
-    run(["golangci-lint run -v"])
+    run(["golangci-lint run -v --disable staticcheck"])
 
 
 def dist():
--- a/do.py
+++ b/do.py
@@ -276,7 +276,7 @@ def go_lint():
     )
     run([pkg])
     os.chdir("pkg")
-    run(["golangci-lint run -v --disable staticcheck"])
+    run(["golangci-lint run -v"])
 
 
 def dist():
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -930,6 +930,9 @@ class Bundler(object):
         for xstatus in self._get_parser("$..x-status").find(self._content):
             status = xstatus.value.get("status")
 
+            if status is not None:
+                status = status.replace("-", "_")
+
             if status not in valid_statuses:
                 raise Exception(
                     "Invalid value for x-status.status={} provided; Valid values are {}".format(
--- a/openapiart/bundler.py
+++ b/openapiart/bundler.py
@@ -921,47 +921,45 @@ class Bundler(object):
                     del content[pieces[-1]]
 
     def _resolve_x_status(self):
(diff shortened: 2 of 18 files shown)


do.py
```diff
@@ -276,7 +276,7 @@ def go_lint():
     )
     run([pkg])
     os.chdir("pkg")
-    run(["golangci-lint run -v"])
+    run(["golangci-lint run -v --disable staticcheck"])
```
Vibaswan, line 279: had to call disable cause otherwise golangci-lint shows usage of deprecation errors,

openapiart/bundler.py
```diff
@@ -598,6 +598,17 @@ def _required_restriction(self, schema_name, value):
                                 property=required, name=schema_name
                             )
                         )
+                    if "x-status" in value["properties"][required].keys():
```
Vibaswan, line 601: added this check as a required property should not have status deprecated, cause it then contradicts with the required nature it-self

do.py
ashutshkumr: isn't there a different approach ? we cannot disable staticcheck altogether

Vibaswan: included a config file for golangci-lint .golangci.yaml which ignores the static check for deprecation
the code ignored is SA1019 more info in https://staticcheck.io/docs/checks#SA1019

dipendughosh (approved):
looks good.

Vibaswan:
models branch name: x-status-warnings
successful models build
https://github.com/open-traffic-generator/models/actions/runs/4312828392/jobs/7523872902


snappi branch name: x-status-warnings
successful snappi build
https://github.com/open-traffic-generator/snappi/actions/runs/4312885711

ashutshkumr:
In generated Go code, the deprecation message should be the very first message.

e.g.

Incorrect:
```go
// Space1 returns a int32
// Deprecated: Information TBD
//
// Description TBD
```

Correct:
```go
// Deprecated: Information TBD
// Space1 returns a int32
//
// Description TBD
```

Need to fix following class of x-status as well.
```go
// UpdateConfig is under Review: the whole schema is being reviewed
//
// Object to Test required Parameter
```

ashutshkumr:
I see disparity in naming conventions in generated Go code. Should be `JB` instead of `J_B`. This is sprinkled throughout.

```go
	if obj.obj.Choice.Number() != 0 {
		enumMap := make(map[string]string)
		enumMap["j_b"] = "J_B enum in property Choice is deprecated, use j_a instead"

		for enum, msg := range enumMap {
			if obj.obj.Choice.String() == enum {
				obj.addWarnings(msg)
			}
		}
	}
```

ashutshkumr (approved):

PR #411, opened 2023-03-17
Error ux

### LINKED ISSUE #359: gRPC responses for equivalent HTTP 4xx or 5xx responses should be returned as an error instead of response object in OK status
**Version information**
v0.2.1

**Describe the bug**
Different response types for HTTP API maps to different protobuf messages with error code in the name of the message.
All those responses are returned only during gRPC OK.

What this implies is, even after doing a check for `err != nil` post gRPC API call, users have to further check for errors inside response messages. This also complicates the retrieval of actual usable response.

Solution is to return all error responses as `err` from gRPC API call. More details TBD.

**To Reproduce**

```go
	request := otg.SetConfigRequest{Config: config.Msg()}
	ctx, cancelFunc := context.WithTimeout(context.Background(), api.grpc.requestTimeout)
	defer cancelFunc()
	resp, err := api.grpcClient.SetConfig(ctx, &request)
	if err != nil {
		return nil, err
	}
	if resp.GetStatusCode_200() != nil {
		return NewResponseWarning().SetMsg(resp.GetStatusCode_200()), nil
	}
	if resp.GetStatusCode_400() != nil {
		data, _ := yaml.Marshal(resp.GetStatusCode_400())
		return nil, fmt.Errorf(string(data))
	}
	if resp.GetStatusCode_500() != nil {
		data, _ := yaml.Marshal(resp.GetStatusCode_500())
		return nil, fmt.Errorf(string(data))
	}
```

**Expected behavior**
```go
	request := otg.SetConfigRequest{Config: config.Msg()}
	ctx, cancelFunc := context.WithTimeout(context.Background(), api.grpc.requestTimeout)
	defer cancelFunc()
	resp, err := api.grpcClient.SetConfig(ctx, &request)

        // should include error responses defined for http 4xx or 5xx codes
	if err != nil {
		return nil, err
	}
```

Vibaswan:
Fixes #359 
Making changes for error-ux for go and python SDKs


Vibaswan:
successful models build
https://github.com/open-traffic-generator/models/actions/runs/4456208500

openapiart/bundler.py
```diff
@@ -366,20 +366,20 @@ def _validate_response_uid(self):
                     )
 
     def _validate_required_responses(self):
-        """Ensure all paths include a 400 and 500 response.
+        """Ensure all paths include a 200 and default response.
 
         Print every path that does not include a 400 or 500 response.
```
ashutshkumr, line 371: this message is not valid anymore

openapiart/bundler.py
```diff
@@ -391,6 +391,23 @@ def _validate_required_responses(self):
                 missing_paths += "{}\n".format(error_message)
         if len(missing_paths) > 0:
             raise Exception(missing_paths)
+
+        # There must be the Error structure in the yaml which should have required kind and errors
```
ashutshkumr, line 395: code is required instead of kind

openapiart/bundler.py
```diff
@@ -366,20 +366,20 @@ def _validate_response_uid(self):
...
+        None: all paths have a 200 and default response
         """
         responses = self._get_parser("$..paths..responses").find(self._content)
-        required_error_codes = ["400", "500"]
+        required_responses = ["200", "default"]
         missing_paths = ""
         for response in responses:
-            missing = set(required_error_codes).difference(
+            missing = set(required_responses).difference(
                 set(response.value.keys())
```
ashutshkumr, line 383: is 200 guaranteed to be string always ? if not, would recommend converting each item in response keys to str

openapiart/bundler.py
```diff
@@ -366,20 +366,20 @@ def _validate_response_uid(self):
...
             )
             if len(missing):
```
ashutshkumr, line 385: i know it's existing code, but check should be `> 0`. Moreover, we don't need to print the error, since we're raising exception already.

openapiart/bundler.py
```diff
@@ -391,6 +391,23 @@ def _validate_required_responses(self):
                 missing_paths += "{}\n".format(error_message)
         if len(missing_paths) > 0:
             raise Exception(missing_paths)
+
+        # There must be the Error structure in the yaml which should have required kind and errors
+        err_schema = self._get_parser("$..Error").find(self._content)
+        if len(err_schema) == 0:
+            raise Exception("default must have a error schema")
```
ashutshkumr, line 398: rephrase: default response must use Error schema

ashutshkumr: BTW, where are we checking that `default` response is pointing to error schema ? if we're not, then validation isn't complete.

I believe currently we're only checking if Error schema exists (in which case I would rephrase the error message above even further)

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: I rephrased the exception a bit , do you have a better suggestion ?

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
ashutshkumr: Multiple concerns:
- We should get a singe match for `Error` as you won't have multiple of them defined; so why aren't we asserting that ? (i.e. exactly one schema MUST exist)
- If "required" is not in error schema, then it's an error (currently it's leaking through)
- Shouldn't we be validating that all default responses actually point to Error schema ?

openapiart/generator.py
```diff
@@ -262,51 +263,36 @@ def _get_methods_and_factories(self):
                     rpc.request_class = class_name
 
             response_type = None
+            proto_name = None
             response_list = operation.get("responses")
             if response_list is None:
                 raise Exception("{} should have responses".format(method_name))
             for response_code, response_property in response_list.items():
-                if int(response_code) == 200:
+                if response_code == "200":
```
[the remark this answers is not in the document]
Vibaswan: done

openapiart/bundler.py
Vibaswan: agreed for first two points, added checks and unit test for the same
for the last point default schema may not directly point to error schema and might have intermediate structures right

openapiart/common.go
```diff
@@ -187,6 +191,54 @@ func (api *api) under_review(message string) {
...
+	obj.Msg().Code = code
+	obj.Msg().Errors = errors
+}
+
+func (api *api) fromGrpcError(err error) (Error, bool) {
+	st, ok := status.FromError(err)
+	if ok {
+		rErr := NewError()
+		if err := rErr.FromJson(st.Message()); err == nil {
+			rErr.Msg().Code = int32(st.Code())
```
Vibaswan, line 219: the code which we got from the server gets overwritten here ?
in this case the code of the rErr will always be 2 which is unknown error

openapiart/common.go
```diff
@@ -187,6 +191,54 @@ func (api *api) under_review(message string) {
...
+			return rErr, true
+		}
+
+		api.setResponseErr(rErr, int32(st.Code()), st.Message())
+		return rErr, true
+	}
+
+	return nil, false
```
Vibaswan, line 227: what if ok is false, I read through the documentation and it says that it returns ok as false if the error is not supported with the package status, now if there is an error which is not supported the package it still gives us st just that ok is false, according to me if ok is false we should still be returning an error otherwise we will miss it

ashutshkumr (approved):

PR #413, opened 2023-03-22
Introduce new data types uint32 and uint64

dipendughosh:


openapiart/generator.py
```diff
@@ -1694,6 +1694,8 @@ def _get_openapi_types(self, yobject):
                 )
                 if min_max > 2147483647:
                     pt.update({key: r"'int64'"})
+                if min_max > 4294967294:
+                    pt.update({key: r"'uint64'"})
```
ashutshkumr, line 1698: I guess we're promoting types here ? Is this only applicable when format is provided ?
I believe it should only be applicable when format is provided.

I would suggest removing this change if this happens even when format has been provided.

Ideally - when format is not provided, we should deduce format based on min-max provided.

openapiart/openapiartgo.py
```diff
@@ -2067,6 +2069,12 @@ def _build_setters_getters(self, fluent_new):
                     and "int" in field.type
                 ):
                     field.type = field.type.replace("32", "64")
+                if (
+                    (field.min is not None and field.min > 4294967294)
+                    or (field.max is not None and field.max > 4294967294)
+                    and "uint" in field.type
+                ):
```
ashutshkumr, line 2076: can we explain this piece of code ? I don't follow why we're checking for both min/max to be greater than this value; also I don't follow mention of uint.

openapiart/openapiartprotobuf.py
```diff
@@ -209,6 +209,23 @@ def _write_header(self, info_object):
         self._write('import "google/protobuf/descriptor.proto";')
         self._write('import "google/protobuf/empty.proto";')
 
+    def _get_integer_format(self, format, min, max):
```
ashutshkumr, line 212: as has been source of my confusion above, we should prioritize checking format first. e.g. here's the step I would expect:
- If format is provided
  - If min-max not compatible with format, raise error
- If format is not provided
  - Based on min-max provided, return appropriate type
  - For this particular release, return int32 if compatible and add a TODO comment to remove it later

ashutshkumr (changes requested):
would request creating branches in models and snappi repo to see how it works there

openapiart/generator.py
dipendughosh: done

openapiart/openapiartprotobuf.py
dipendughosh: done

PR #415, opened 2023-04-05
modifying python and go files for error-ux modification

Vibaswan:
modification in snappi for proper alignment of error reporting
changes in generated controller in gosnappi to return error obj

### DIFF
--- a/openapiart/generator.py
+++ b/openapiart/generator.py
@@ -415,30 +415,6 @@ class Generator:
             err.errors = [grpc_error.details()]
         raise Exception(err)
 
-    def from_exception(self, grpc_error):
-        # type: (Exception) -> Union[Error, None]
-
-        if isinstance(grpc_error, Error):
-            return grpc_error
-        elif isinstance(grpc_error, grpc.RpcError):
-            err = self.error()
-            err.code = grpc_error.code().value[0]
-            err.errors = [grpc_error.details()]
-            return err
-        else:
-            if len(grpc_error.args) != 1:
-                return None
-
-            if isinstance(grpc_error.args[0], Error):
-                return grpc_error.args[0]
-            elif isinstance(grpc_error.args[0], str):
-                err = self.error()
-                try:
-                    err.deserialize(grpc_error.args[0])
-                except Exception as _:
-                    return None
-                return err
-
     @property
     def request_timeout(self):
         \"\"\"duration of time in seconds to allow for the RPC.\"\"\"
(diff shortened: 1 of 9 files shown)


Vibaswan:
models and snappi branch names `error-ux-modify`
successful model repo CI pass 
https://github.com/open-traffic-generator/models/actions/runs/4620277191/jobs/8170150040
successful snappi repo CI pass
https://github.com/open-traffic-generator/snappi/actions/runs/4620424190/jobs/8170504877

openapiart/generator.py
```diff
@@ -406,17 +406,38 @@ def _serialize_payload(self, payload):
...
             err = self.error()
-            try:
-                err.deserialize(grpc_error.details())
-            except Exception as e:
-                err.code = 2 # code for unknown error
-                err.errors = [grpc_error.details()]
-        return err
+            err.code = grpc_error.code().value[0]
+            err.errors = [grpc_error.details()]
+            return err
```
ashutshkumr, line 427: Why aren't we deserializing `grpc_error.details()` to `error` ?

openapiart/generator.py
```diff
@@ -406,17 +406,38 @@ def _serialize_payload(self, payload):
...
+        else:
+            if len(grpc_error.args) != 1:
```
ashutshkumr, line 429: shouldn't we check if `args` is even a valid property ? (since we're not always expecting an Exception object)

openapiart/generator.py
```diff
@@ -406,17 +406,38 @@ def _serialize_payload(self, payload):
...
+    def _raise_exception(self, grpc_error):
+        err = self.error()
+        try:
+            err.deserialize(grpc_error.details())
+        except Exception as _:
+            err.code = grpc_error.code().value[0]
+            err.errors = [grpc_error.details()]
+        raise Exception(err)
+
     def from_exception(self, grpc_error):
```
ashutshkumr, line 418: calling it `grpc_error` is a misnomer

openapiart/goserver/go_controller_generator.py
```diff
@@ -278,7 +278,8 @@ def _write_method(self, w, ctrl, route):
                     result = {models_prefix}NewError()
                     err := result.FromJson(rsp_err.Error())
                     if err != nil {{
-                        result = nil
+                        result.Msg().Code = int32(status_code)
```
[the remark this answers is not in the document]
ashutshkumr: we should never assign kind automatically - if client is not receiving kind (when it's expected), then most we need to ensure that server is setting it.

ashutshkumr (changes requested):
I find it concerning now that in python, we have same kind of if checks repeated mulitple times just because we handle error differently for grpc and http.

I don't quite like it and it has resulted in more than 10 different possible outcomes.
Hence I would request to see if there's an chance of simplifying from_exception, raise_exception, etc.

openapiart/goserver/go_controller_generator.py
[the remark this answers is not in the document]
Vibaswan: its regarding the point Biplab mentioned in the email if request is nil or an error occurred while parsing json and stuff , technically this code is also server side code right ?

openapiart/generator.py
Vibaswan: done

openapiart/generator.py
Vibaswan: changed it

openapiart/generator.py
Vibaswan: done

ashutshkumr (approved):

PR #419, opened 2023-04-20
changes for minimize requirements in python and handle pipreqs issue

Vibaswan:


ashutshkumr (approved):
Some more changes needed in how we generate requirements. Needs to be addressed as a separate PR.

PR #427, opened 2023-06-14
fixing error for property with name string

Vibaswan:


openapiart/openapiartgo.py
```diff
@@ -349,7 +349,10 @@ def _get_external_field_name(self, openapi_name):
         return external
 
     def _get_external_struct_name(self, openapi_name):
-        return self._get_external_field_name(openapi_name).replace("_", "")
+        struct_name = self._get_external_field_name(openapi_name).replace(
```
Vibaswan, line 352: changing property name String() to StringVal(), we can think of other names to, cannot have String() as we already use it for for getting the string output of our gosnappi object

Vibaswan:
closing this as this is not needed anymore

Issue #424, opened 2023-06-05
Any 2xx response should satisfy the response requirements for openapi requests


pmajumdarKS:
In HTTP, 200 is a response code which normally comes with a body.

Response code 204 means that the response doesn't have any content. This is often the most appropriate response code for the HTTP DELETE requests.

Similarly 201 is a more appropriate response for POST requests.

Forcing to add a 200 response code for all requests doesn't make sense. Any 2xx should suffice.

PR #428, opened 2023-06-15
[Issue #424] [Potential solution] Any 2xx response should be good eno…

pmajumdarKS:
…ugh. It doesn't have to be 200.


ashutshkumr:
We do not plan to support multiple varieties of successful responses for a single HTTP request

PR #429, opened 2023-06-16
Fix python lint check

Vibaswan:
python do.py lint now formats the python files using black and provides information on which files changed and remained same on the out 
CI uses python do.py true (check=true) and prints all the files that needs reformatting and remained unchanged 
example CI failure due to lint issues 
https://github.com/open-traffic-generator/openapiart/actions/runs/5286727208/jobs/9566426810#step:6:17

Pass CI also prints that all files are good and no of unchanged files
https://github.com/open-traffic-generator/openapiart/actions/runs/5286810785/jobs/9566599826#step:6:9

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
           python do.py install
       - name: Run python lint
         run: |
-          python do.py lint
+          python do.py lint true
       - name: Archive generated artifacts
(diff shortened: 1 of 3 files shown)


ashutshkumr:
@pmajumdarKS FYI, this is addressed.

ashutshkumr (approved):

PR #431, opened 2023-06-16
Dev workflow

### LINKED ISSUE #430: Dev workflow is currently very painful
image

Vibaswan:
Fixes #430

### DIFF
--- a/do.py
+++ b/do.py
@@ -281,6 +281,7 @@ def go_lint():
     run([pkg])
     os.chdir("pkg")
     run(["golangci-lint run -v"])
+    os.chdir("..")
 
 
 def dist():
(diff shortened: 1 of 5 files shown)


do.py
```diff
@@ -459,6 +460,42 @@ def getstatusoutput(command):
     )
 
 
+def build(clean_env="false", sdk="all"):
+    print("\nStep 1: Set up virtaul env, if does not exsists already")
```
ashutshkumr, line 464: remove `if does not *` since you're already printing the messages below

do.py
```diff
@@ -459,6 +460,42 @@ def getstatusoutput(command):
     )
 
 
+def build(clean_env="false", sdk="all"):
```
ashutshkumr, line 463: sdk should be the first argument as that's more frequently going to be used.

clean_env accepting false/true is not intuitive; maybe use `env_setup` with possible values as None and "clean"; we might add more later (e.g. reinstall_deps, etc.)

do.py
Vibaswan: done

do.py
Vibaswan: done

do.py
```diff
@@ -459,6 +460,42 @@ def getstatusoutput(command):
...
+    if clean_env.lower() == "true":
+        print("\nCleaning up exsisting env")
+        run(["rm -rf .env"])
+
+    if not os.path.exists(".env"):
+        setup()
+    else:
+        print("virtualenv already exists will be the exsisting one\n")
+
+    py.path = os.path.join(".env", "bin", "python")
```
ashutshkumr, line 477: Print that now we'll be using following interpreter path.

ashutshkumr (changes requested):
Added all review comments

ashutshkumr (changes requested):
Also, for Dockerfile, why are we only generating python SDK ?

ashutshkumr (approved):

PR #433, opened 2023-07-18
Data type refactor restructuring and changes

### LINKED ISSUE #436: valid formats for supported integer type is not checked
if a integer type is missing format specify warning
if an unknown format is specified apart from int32, int64, uint32, uint64 of type integer error must be thrown

Vibaswan:
Fixes #436 , #437 and #438


openapiart/bundler.py
```diff
@@ -93,6 +93,7 @@ def bundle(self):
             self._read_file(self._base_dir, self._api_filename)
 
         self._resolve_x_include()
+        self._validate_x_field_pattern()
```
ashutshkumr, line 96: should be called inside `_resolve_x_pattern`

also, is there a reason why we pass a value "x-field-pattern" ? wouldn't it just make sense to call this method `_resolve_x_field_pattern` ?

openapiart/bundler.py
```diff
@@ -635,6 +663,37 @@ def _required_restriction(self, schema_name, value):
...
+                    % (
+                        str(xpattern_path.full_path),
+                        xpattern["format"],
+                        str(valid_formats),
+                    )
+                )
+        if len(self._errors) > 0:
+            self._validate_errors()
+
     def _resolve_x_pattern(self, pattern_extension):
```
ashutshkumr, line 697: can we replace all variables called `format` with `fmt` ? it's a keyword in python

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
```diff
@@ -635,6 +663,37 @@ def _required_restriction(self, schema_name, value):
...
+        ):
+            xpattern = xpattern_path.value
+
+            if "format" not in xpattern:
+                self._errors.append(
+                    "{} is not valid, format is mandatory for x-field-pattern".format(
+                        str(xpattern_path.full_path)
+                    )
+                )
+            elif xpattern["format"] == "integer" and "length" not in xpattern:
```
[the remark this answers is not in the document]
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

ashutshkumr:
Do we need this snippet ? Why ?

```
        if "enums" in xpattern:
            schema["properties"]["value"]["enum"] = copy.deepcopy(
                xpattern["enums"]
            )
            schema["properties"]["values"]["items"]["enum"] = copy.deepcopy(
                xpattern["enums"]
```

ashutshkumr:
Do we need this too ? can't it be deepcopied directly to schema ?

```
        xconstants = (
            xpattern["x-constants"] if "x-constants" in xpattern else None
        )
```

ashutshkumr:
This is again coming from old code, but shouldn't we check if `counter` feature is provided before defining Incr/Decr schemas ?

Vibaswan:
these were from beginning not sure the use-cases

ashutshkumr:
> these were from beginning not sure of the use-cases

Would suggest removing any unused code, hence asking for confirmation if it's indeed used

openapiart/bundler.py
```diff
@@ -939,18 +1013,27 @@ def _apply_common_x_field_pattern_properties(
...
+        finalised_format = None
+        if xpattern["format"] == "integer":
+            finalised_format = "uint32"
+            if "length" in xpattern and xpattern["length"] > 32:
+                finalised_format = "uint64"
+        elif fmt is not None:
+            finalised_format = fmt
+
+        if finalised_format is not None:
             # TODO: fix this
```
ashutshkumr, line 1032: remove todo

openapiart/bundler.py
```diff
@@ -939,18 +1013,27 @@ def _apply_common_x_field_pattern_properties(
...
-            # if property_name == "values":
-            #     schema["items"]["format"] = format
-            # else:
-            schema["format"] = format
+            if property_name == "values":
+                schema["items"]["format"] = finalised_format
+            else:
+                schema["format"] = finalised_format
         if "length" in xpattern:
             # TODO: fix this
```
ashutshkumr, line 1038: remove stale comments

openapiart/bundler.py
```diff
@@ -939,18 +1013,27 @@ def _apply_common_x_field_pattern_properties(
...
-            # if property_name == "values":
-            #     schema["items"]["format"] = format
-            # else:
-            schema["format"] = format
+            if property_name == "values":
+                schema["items"]["format"] = finalised_format
+            else:
+                schema["format"] = finalised_format
         if "length" in xpattern:
```
ashutshkumr, line 1037: when are we setting min/max for items in values ?

ashutshkumr:
We're not setting correct format and min/max `auto` field (for `integer`)

openapiart/bundler.py
```diff
@@ -884,6 +956,7 @@ def _generate_value_schema(
                     "offset": {
                         "description": "Offset in bits relative to start of corresponding header field",
                         "type": "integer",
+                        "format": "uint32",
```
ashutshkumr, line 959: why is this hardcoded ? shouldn't it depend on bit length ?

openapiart/bundler.py
```diff
@@ -892,6 +965,7 @@ def _generate_value_schema(
                     "length": {
                         "description": "Number of bits to track for metrics starting from configured offset of corresponding header field",
                         "type": "integer",
+                        "format": "uint32",
```
ashutshkumr, line 968: why is this hardcoded ? shouldn't it depend on bit length ?

openapiart/bundler.py
```diff
@@ -833,22 +903,24 @@ def _generate_value_schema(
                     },
                 },
             }
+
             if "features" in xpattern and "count" in xpattern["features"]:
                 counter_schema["properties"]["count"] = {
                     "type": "integer",
+                    "format": "uint32",
```
ashutshkumr, line 910: why is this hardcoded ? shouldn't it depend on bit length ?

ashutshkumr:
Do not provide minimum and maximum for properties which require exact 32/64 bits.

openapiart/common.py
```diff
@@ -423,14 +423,25 @@ def validate_hex(self, hex):
...
+                return False
+            elif type_format == "uint64" and (
+                value < 0 or value > 18446744073709551615
+            ):
+                return False
+            elif type_format == "int32" and (
+                value < -2147483648 or value > 2147483647
+            ):
+                return False
+            elif type_format == "int32" and (
```
ashutshkumr, line 441: should be "int64" (maybe add more tests since this wasn't caught ?)

openapiart/common.py
```diff
@@ -423,14 +423,25 @@ def validate_hex(self, hex):
...
             return False
-        if type_format == "uint32" and value < 0:
-            return False
-        if type_format == "uint64" and value < 0:
-            return False
         if min is not None and value < min:
             return False
         if max is not None and value > max:
             return False
+        if type_format is not None:
```
ashutshkumr, line 430: if format is not mentioned, it should be assumed to be int32 - don't see any code for checking that (and the min-max range)

openapiart/common.py
```diff
@@ -903,11 +914,17 @@ def _validate_types(self, property_name, property_value):
             msg = "property {} shall be of type {} at {}".format(
                 property_name, details["type"], self.__class__
             )
+
+            itemtype = (
```
ashutshkumr, line 918: without this change, were we seeing any bugs ?

Vibaswan: yes the the validation was not happening for list of type integers

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

openapiart/bundler.py
Vibaswan: done

Vibaswan:
> > these were from beginning not sure of the use-cases
> 
> Would suggest removing any unused code, hence asking for confirmation if it's indeed used

removed unused code

openapiart/common.py
Vibaswan: done

openapiart/common.py
Vibaswan: added it here in generator.py line number 1754

ashutshkumr:
empty string not needed
```
                    pt.get("type", "") == "int" or pt.get("itemtype") == "int"

```

ashutshkumr:
Inside `_get_integer_format`, if `type_format` is None, set it to `int32`

ashutshkumr:
needs to be `and`

```
    def _min_max_in_range(cls, format, min, max):
        if min is None or max is None:
            return True
```

ashutshkumr:
Remove `_format_from_range` and its usage. Print a WARNING that since format has not been provided, it's being defaulted to `int32`

openapiart/common.py
ashutshkumr: set default `type_format` to `int32`

ashutshkumr:
Probably need to include map for `int32`
```
self._oapi_go_types = {
```

PR #445, opened 2023-09-11
Commit sdk to repo

Vibaswan:
changes to commit python and go gen files to repo


ashutshkumr (changes requested):
can we please steer off from name `sanity` ?

1. We should use a package name `refapi` for go, python and protobuf. For protobuf SDKs, it would use `refapipb`
2. When package name is not provided, we should just use `oapi` instead of `sanity` as defaults

Vibaswan:
> can we please steer off from name `sanity` ?
> 
> 1. We should use a package name `refapi` for go, python and protobuf. For protobuf SDKs, it would use `refapipb`
> 2. When package name is not provided, we should just use `oapi` instead of `sanity` as defaults

Have done the changes accordiongly

PR #455, opened 2023-10-18
Refactor go sdk

Vibaswan:
1.  All public methods for creation of structs will be removed from GosnappiApi interface
     a.	Those public methods will only be accessible on package level (e.g. gosnappi.NewConfig() will be valid api.NewConfig() won’t be valid)
    b.	Api interface will be made private and GosnappiApi will be renamed to Api
    c.	FromError() method will be exposed on package level instead of api level
    d.	Warnings() method will be made private, until we decide how we really want to use it
2. There are helper methods defined on each struct which we’ll reorganize or hide
    a.	Msg(), SetMsg() and Validate() will now be private
    b.	Clone() and String() will remain as is
    c.	All marshalling methods will be moved behind a single method SerDes().
    e.g. ob.ToJson() or ob.FromJson() will now be replaced by ob.SerDes().ToJson() or ob.SerDes().FromJson()
3.	Choice setter SetChoice() was leading to “two ways of doing the same thing” and hence will be made private
a.	Since in gosnappi, choice is supposed to be implicitly set based on the choice property set by the user

Successful snappi CI run
https://github.com/open-traffic-generator/snappi/actions/runs/6557775808
Note: some of the gonappi tests seems not valid after the changes so commented those out


ashutshkumr:
Check review comments at https://github.com/open-traffic-generator/snappi/pull/214

ashutshkumr (approved):

PR #456, opened 2023-11-29
adding support for signed integers in x-field-pattern

Vibaswan:
Adding a new property in x-field-pattern
```yaml
signed:
  description: >-
    Valid only if the format is of type integer.
    Indicates whether the integer value is signed or unsigned.
    If the property is not specified the value is considered as unsigned.
    Its a boolean value which is optional in nature.
  type: boolean
```

Example model snippet (`Users snippet`)
```yaml
SignedIntegerPattern:
  description: Test signed integer pattern
  type: object
  properties:
    integer:
      x-field-pattern:
        format: integer
        default: 0
        signed: true
        length: 8
        features: [count]
      x-field-uid: 1
```

Final output in yaml after bundling
```yaml
Pattern.SignedIntegerPattern.Integer:
      description: |-
        TBD
      type: object
      properties:
        choice:
          type: string
          x-enum:
            value:
              x-field-uid: 2
            values:
              x-field-uid: 3
            increment:
              x-field-uid: 4
            decrement:
              x-field-uid: 5
          default: value
          x-field-uid: 1
          enum:
          - value
          - values
          - increment
          - decrement
        value:
          type: integer
          x-field-uid: 2
          default: 0
          format: int32
          minimum: -256
          maximum: 255
        values:
          type: array
          items:
            type: integer
            format: int32
            minimum: -256
            maximum: 255
          x-field-uid: 3
          default:
          - 0
        increment:
          $ref: '#/components/schemas/Pattern.SignedIntegerPattern.Integer.Counter'
          x-field-uid: 5
        decrement:
          $ref: '#/components/schemas/Pattern.SignedIntegerPattern.Integer.Counter'
          x-field-uid: 6
```


MODELGUIDE.md
```diff
@@ -181,6 +181,13 @@ x-field-pattern:
       Pre-processing will write minimum and maximum values based on the length.
       Length will be ignored for mac, ipv4, ipv6 formats.
     type: integer
+  signed:
+    description: >-
+      Valid only if the format is of type integer or checksum.
```
ashutshkumr, line 186: checksums are never signed

openapiart/bundler.py
```diff
@@ -683,6 +683,22 @@ def _validate_x_field_pattern(self, xpattern_path):
                     str(xpattern_path.full_path)
                 )
             )
+        if "signed" in xpattern:
+            if (
+                xpattern["format"] != "integer"
+                and xpattern["format"] != "checksum"
```
ashutshkumr, line 689: why do we need to validate this for checksum ?

openapiart/bundler.py
```diff
@@ -1047,9 +1062,18 @@ def _apply_common_x_field_pattern_properties(
             else:
                 schema["format"] = finalised_format
         if "length" in xpattern and int(xpattern["length"]) not in [32, 64]:
+            min_val = (
```
ashutshkumr, line 1065: for unsigned, range: 0 to (2^N) - 1
for signed, range: -2^(N-1) to 2^(N-1)-1

where N is number of bits or length

ashutshkumr (changes requested):

ashutshkumr (approved):
LGTM

PR #467, opened 2024-06-12
add support for random UDF for x-field-pattern fields

Vibaswan:
**Introducing New random feature for x-field-pattern**

Random for Ipv4 Pattern
```yml
# initial definition
ipv4:
# other parameters
 dst:
  x-field-pattern:
    format: ipv4
    default: 0.0.0.0
    features: [random]
  x-field-uid: 1

# after expansion by openapiart
Pattern.Ipv4Pattern.Ipv4.Dst:
  type: object
  properties:
    choice:
      type: string
      x-enum:
        value:
          x-field-uid: 2
        values:
          x-field-uid: 3
        random:
          x-field-uid: 6
      default: value
      x-field-uid: 1
      enum:
      - value
      - values
      - random
    value:
      type: string
      x-field-uid: 2
      default: 0.0.0.0
      format: ipv4
    values:
      type: array
      items:
        type: string
        format: ipv4
      x-field-uid: 3
      default:
      - 0.0.0.0
    random:
     $ref: '#/components/schemas/Pattern.Ipv4Pattern.Ipv4.Dst.Random'
     x-field-uid: 4

Pattern.Ipv4Pattern.Ipv4.Dst.Random:
  description: |-
    ipv4 random pattern
  type: object
  properties:
    min:
      type: string
      x-field-uid: 1
      default: 0.0.0.0
      format: ipv4
    max:
      type: string
      x-field-uid: 2
      default: 255.255.255.255
      format: ipv4
    seed:
      type: integer
      x-field-uid: 3
      default: 1
      format: uint32
    count:
      type: integer
      x-field-uid: 4
      default: 1
      format: uint32
```

Random for mac pattern
```yml
# initial definition
mac:
# other parameters
 dst:
  x-field-pattern:
    format: mac
    features: [random]
  x-field-uid: 1

# after expansion by openapiart
# skipping intermediate strcuture

Pattern.MacPattern.Mac.Random:
  description: |-
    mac random pattern
  type: object
  properties:
    min:
      type: string
      x-field-uid: 1
      default: 00:00:00:00:00:00
      format: mac
    max:
      type: string
      x-field-uid: 2
      default: ff:ff:ff:ff:ff:ff
      format: mac
    seed:
      type: integer
      x-field-uid: 3
      default: 1
      format: uint32
    count:
      type: integer
      x-field-uid: 4
      default: 1
      format: uint32
```

Random for integer pattern
```yml
# initial definition
integer_pattern:
# other parameters
 integer:
  x-field-pattern:
    format: integer
    default: 0
    length: 8
    features: [random]
  x-field-uid: 1

# after expansion by openapiart
# skipping intermediate structure

Pattern.IntegerPattern.Integer.Random:
  description: |-
    integer random pattern
  type: object
  properties:
    min:
      type: integer
      x-field-uid: 1
      default: 0
      format: uint32
      maximum: 255
    max:
      type: integer
      x-field-uid: 2
      default: 255
      format: uint32
      maximum: 255
    seed:
      type: integer
      x-field-uid: 3
      default: 1
      format: uint32
      maximum: 255
    count:
      type: integer
      x-field-uid: 4
      default: 1
      format: uint32
      maximum: 255
```

**sample python code to use random**
```python
random = config.ipv4.dst.random
random.min = "1.2.3.4"
random.max = "255.255.3.4"
random.count = 2
random.seed = 2
```

**sample go code to use random**
```go
r := Config.Ipv4().Dst().Random()
r.SetMin("1.2.3.4").SetMax("255.255.3.4").SetCount(2).SetSeed(2)
```


openapiart/bundler.py
```diff
@@ -1060,6 +1060,68 @@ def _generate_value_schema(
...
+                            "type": type_name,
+                            "x-field-uid": random_auto_field.uid,
+                        },
+                        "max": {
+                            "description": "The maximum possible value generated by the random value generator.",
+                            "type": type_name,
+                            "x-field-uid": random_auto_field.uid,
+                        },
+                        "seed": {
+                            "description": "The seed value to initialize the random value generator to a deterministic state.",
```
lthompson-keysight, line 1093: I would suggest that we default the seed value to 0, but treat this as a special case which causes the seed to be set to a random value. This would mimic the behavior of random functions in software libraries. But it's not crucial since the script can always generate a random number to use

openapiart/bundler.py
```diff
@@ -1060,6 +1060,68 @@ def _generate_value_schema(
                 metric_tags_schema_name
             ] = metric_tags_schema
 
+        if "features" in xpattern and "random" in xpattern["features"]:
+            if xpattern["format"] in ["integer", "ipv4", "ipv6", "mac"]:
+                random_pattern_name = "{}.Random".format(schema_name)
```
lthompson-keysight, line 1065: Should we specify the behavior of the implementation for cases where the range of random values is less than the range of the requested format? In particular some implementations may not be able to randomize the whole ipv6 field.

openapiart/bundler.py
```diff
@@ -1060,6 +1060,68 @@ def _generate_value_schema(
...
+                    )
+                }
+                schema["properties"]["random"]["x-field-uid"] = auto_field.uid
+                random_auto_field = AutoFieldUid()
+                random_schema = {
+                    "description": "{} random pattern".format(
+                        xpattern["format"]
+                    ),
+                    "type": "object",
+                    "properties": {
```
lthompson-keysight, line 1081: Should we have options to exclude loopback, link-local, and/or mutlicast addresses when either IP format is chosen?

openapiart/bundler.py
Vibaswan: So should be specify that in the description of ipv6 fields in random ?
what should we specify as max for ipv6 ?

openapiart/bundler.py
Vibaswan: well this is a generic Idea should not the person building model be aware and not include those fields with random support ?

openapiart/bundler.py
PrasenjitAdhikary: If range of random values is less than the requested format, then implementation will raise error to reduce requested range. We can add this support once we need to support IPv6 / other format, where such situation can arise.

openapiart/bundler.py
PrasenjitAdhikary: If we have such requirement, model can be extended to add such exclusion list without breaking backward compatibility. openapiart can be extended to support such exclusion specific to a field format. Assumption, backend implementation (UHD/OTG) can also support such exclusion.

openapiart/bundler.py
PrasenjitAdhikary: We will update model description to treat 0 as special seed value (which will be replaced by a random value in the implementation to get true randomness). But we want to set default seed value to 1 so that test cases by default generates predictable sequence, which is easy to debug, for any failure. The user can opt for a seed value of 0 if they want different behavior.

openapiart/bundler.py
lthompson-keysight: A range of less than the requested format is a valid and common use case, so we must allow it. But we should change min and max to accommodate 128-bit min and max values. For UHD, the range of random numbers (max-min+1) must fit within a uint32, but we could conceivably allow a 128-bit min value, though we are currently implemented as a 32-bit min value.

lthompson-keysight: But since this review only affects openapiart it's fine for now.

lthompson-keysight (approved):

MODELGUIDE.md
```diff
@@ -224,7 +224,7 @@ Flow.Ipv4:
       x-field-pattern:
         format: ipv4
         default: 0.0.0.0
-        count: false
+        features: [count, random]
```
alakendu, line 227: According to docs `x-device-pattern` has similar features. So, `random` should applicable for `x-device-pattern` though I can't find this `x-device-pattern` pattern within test. Please extend test coverage if `x-device-pattern` present within bundler code.

Vibaswan: I dont think we use x-device pattern any more , its been removed

openapiart/tests/pattern/pattern.yaml
```diff
@@ -8,7 +8,7 @@ components:
           x-field-pattern:
             format: ipv4
             default: 0.0.0.0
-            features: [count]
+            features: [count, random]
```
alakendu, line 11: Add one test without `random` feature and validate that `random` will not auto-populate 
```
        ipv4:
          x-field-pattern:
            format: ipv4
            default: 0.0.0.0
            features: [count, auto]
```

Vibaswan: can this be treated as an example , test is already present to check this pattern 
https://github.com/open-traffic-generator/openapiart/pull/467/files#diff-7df96b39be93404f05e016f7517caf3a05521ba323a82e6be906138fb72edcfcR87

openapiart/bundler.py
```diff
@@ -863,224 +863,323 @@ def _generate_value_schema(
...
+            if xconstants is not None:
+                counter_schema["x-constants"] = copy.deepcopy(xconstants)
+
             self._content["components"]["schemas"][
-                metric_tags_schema_name
-            ] = metric_tags_schema
+                counter_pattern_name
+            ] = counter_schema
 
-        # TODO; temporary hack to add auto hierarchy field at the end, remove later
```
alakendu, line 1064: make sure we cover following scenario before removing TODO
1. Test should not regress, expecting you have some test.
2. I couldn't figure out proper fix of this todo. Could you please pointing out that change.

MODELGUIDE.md
```diff
@@ -224,7 +224,7 @@ Flow.Ipv4:
       x-field-pattern:
```
alakendu, line 224: `random` must be incorporate in the definition of `x-field-pattern` . I am pointing this help section:
```
  features:
    type: string
    enum: [count, auto, metric_group]
    description: >- 
      count:
      Used to specify .....
```
would eb something :
```
  features:
    type: string
    enum: [count, auto, random, metric_group]
    description: >- 
      count:
      Used to specify 
      ....
      random:
      <defined random>
```

openapiart/tests/pattern/pattern.yaml
```diff
@@ -18,7 +18,7 @@ components:
           x-field-pattern:
             format: ipv6
             default: "::"
```
alakendu, line 20: Check without default Ipv6
```
        ipv6:
          x-field-pattern:
            format: ipv6
            features: [count, random]
          x-field-uid: 1
```

openapiart/bundler.py
```diff
@@ -1091,17 +1190,24 @@ def _apply_common_x_field_pattern_properties(
             "ipv4": "0.0.0.1",
             "ipv6": "::1",
         }
+        max_defaults = {
+            "mac": "ff:ff:ff:ff:ff:ff",
+            "ipv4": "255.255.255.255",
```
alakendu, line 1195: What about `ipv6`

MODELGUIDE.md
Vibaswan: already mentioned here , will add to openapiart as well
https://github.com/open-traffic-generator/models/pull/380/files#diff-174ecc7450d75d1a117f691cdbe9b658d8843cf43d1177c91548995156a9ab6d

openapiart/bundler.py
Vibaswan: Earlier the fix was to add new auto parameters to end by default forcefully , now the solution is that the features are added according to their position in the features list

pkg/unit1_test.go
```diff
@@ -26,3 +26,48 @@ func TestToAndFromProtoMsg(t *testing.T) {
...
+
+	pat2 := config2.Ipv4Pattern().Ipv4().Random()
+	assert.Equal(t, pat2.Min(), "0.0.0.0")
+	assert.Equal(t, pat2.Max(), "255.255.255.255")
+	assert.Equal(t, pat2.Count(), uint32(1))
+	assert.Equal(t, pat2.Seed(), uint32(1))
+
+	pat3 := config2.Ipv6Pattern().Ipv6().Random()
+	assert.Equal(t, pat3.Min(), "::")
+	assert.Equal(t, pat3.Max(), "::")
```
alakendu, line 64: is this expected `Max` value for IPv6 (e,g, `::`)

MODELGUIDE.md
alakendu: ok resolved it. Please note if we extended that document to defined what type of  `properties` we have for `random`. yes, same thing is missing for `increment` and `decrement`.

openapiart/tests/pattern/pattern.yaml
Vibaswan: done

pkg/unit1_test.go
Vibaswan: corrected it

openapiart/bundler.py
Vibaswan: added max value for ipv6

alakendu (approved):

PR #494, opened 2025-10-15
Issue grpc anchor

lthompson-keysight:
Anchor to newer protobuf and grpc for python 3.10+
- protobuf~=5.29.0 ; python_version >= '3.10'
- grpc*~=1.70.0 ; python_version > '3.6' and python_version < '3.10'


lthompson-keysight:
I could bump the the conditional requirement to python 3.11 if there are issues with other projects using the new anchors with python 3.8 through 3.10. May even be able to go to 3.12 for Cobalt, but would require some further investigation.

PR #507, opened 2026-05-05
Preserve security definitions in bundled OpenAPI output

Vibaswan:
**Summary**
- The bundler was silently dropping the root-level security field and components.securitySchemes when bundling multi-file OpenAPI specs. This PR fixes that.
- Added "security" to the list of top-level keys copied from source YAML files in _process_yaml_object.
- Initialized securitySchemes: {} in the default components structure and added name validation for security schemes, consistent with how schemas and responses are handled.
- Fixed a crash in _check_nested_components when a components block contains only securitySchemes and no schemas key.

**Test plan**

-  Added openapiart/tests/security/security.yaml — a fixture covering bearer, API key, OAuth2, and OpenID Connect schemes, plus root-level and operation-level security overrides.
-  Added openapiart/tests/test_security_definitions.py — verifies that after bundling, the output YAML contains the correct root security list, all four securitySchemes entries with their full structure, and per-operation security overrides (/status with OAuth2 scope, /public with empty list).

**_Note:_** `no validation for the security schemas has been added `

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        python-version: [3.9]
+        python-version: [ 3.9 ]
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


lthompson-keysight:
LGTM

lthompson-keysight (approved):
LGTM

